PR #69, opened 2026-04-23
Audit: knowledge-graph findings — Known_Providers refactor + multilingual edge tests + REST category payload

fabiodalez-dev:
## Summary

Branch produced by a dual knowledge-graph audit (**graphify** semantic + **code-review-graph** Tree-sitter) run on 2026-04-22. Captures the three concrete improvements the audit surfaced plus one pre-existing banner-rest work-in-progress that was sitting uncommitted on `main`.

- **refactor**: `Known_Providers::get_all()` (1861 lines → 91) moved to `includes/data/known-providers.json`, lazy-loaded with a per-request static cache. 143 services, 565 URL patterns, 212 cookie patterns — byte-equal output to the previous literal, verified against all 12 callers.
- **test**: 4 new Playwright specs in `tests/e2e/specs/multilingual-edge.spec.ts` that cover the cache-safety chokepoints (Vary header emission, `_fazResolveBrowserLanguage` region-strip/no-match/no-op paths). Locally green in 14.8s.
- **feat (banner-rest)**: `/faz/v1/banner/{lang}` now exposes full category metadata (id, name, description, isNecessary, ccpaDoNotSell, defaultConsent, per-category cookies) and filters out `wordpress-internal` server-side.
- **docs**: graph-audit baseline (`docs/blast-radius-core.md`) + upstream bug tracker for [tirth8205/code-review-graph#372](https://github.com/tirth8205/code-review-graph/issues/372) (three MCP tools raise `'str'/'NoneType' has no attribute resolve'`).

## Why this matters

| Change | Concrete win |
|---|---|
| JSON data dump | Opcache no longer loads a 1861-line literal on every cold autoload; editing the provider map stops triggering PHP re-parse |
| Multilingual tests | Plugs the only untested surface of PR #68 (issue #67): `send_vary_header`, `_fazResolveBrowserLanguage`, `_fazMaybeSwapLanguage` |
| Full category payload | Removes a class of bugs where the frontend banner missed `defaultConsent`, `cookies`, `isNecessary` flags after a REST language swap |

## Test plan

- [x] `npm run test:e2e -- multilingual-edge.spec.ts` → **4/4 green (14.8s)**
- [x] `php -l includes/class-known-providers.php` → no syntax errors
- [x] PHP equivalence check: `Known_Providers::get_all()` returns 143 entries with identical structure pre/post refactor; `get_cookie_map()` and `get_pattern_map()` unchanged
[...]

PR #70, opened 2026-04-23
test: deterministic swap marker + shared wpCreds fixture (follow-up to #69)

fabiodalez-dev:
## Summary

Addresses the two CodeRabbit nitpick comments on PR #69 (`multilingual-edge.spec.ts:115` and `:40-61`) that were intentionally deferred because they touched surfaces outside the original audit scope (frontend JS + fixture infrastructure).

**Stacked PR**: base = `audit/post-graph-improvements`. When #69 merges to `main`, GitHub will automatically re-target this PR's diff to `main`.

### 1. `_fazStore._swapResolved` deterministic marker
`frontend/js/script.js` — `_fazInit()` now flips `_fazStore._swapResolved = true` in a `finally` block around `_fazMaybeSwapLanguage()`. Because `_fazStore` is an alias of `window._fazConfig` (script.js:4), the flag is read back through `window._fazConfig._swapResolved`. Additive, back-compatible, zero behaviour change for production callers.

Regenerated `frontend/js/script.min.js` via `npm run build:min` per project policy.

### 2. Worker-scoped `wpCreds` fixture
`tests/e2e/fixtures/wp-fixture.ts` gains a worker-scoped `wpCreds: { baseURL, adminUser, adminPass, deployPath }` fixture. Existing test-scoped `wpBaseURL`, `adminUser`, `adminPass` now re-export from `wpCreds` so any spec using them already gets the single source of truth; specs that need the values inside `beforeAll`/`afterAll` (which can't consume test-scoped fixtures) can now destructure `wpCreds` directly.

### 3. Consumers migrated
`multilingual-edge.spec.ts` and `multilingual-cache.spec.ts` now:
- consume `wpCreds` inside `beforeAll`/`afterAll` instead of re-reading `process.env` with hardcoded defaults
- (edge spec only) replace `page.waitForTimeout(750)` with `page.waitForFunction(() => window._fazConfig?._swapResolved === true)` on the no-op-swap test, so slow CI runners can no longer false-green by racing an unresolved fetch

## Test plan

```
$ npm run test:e2e -- multilingual-edge.spec.ts multilingual-cache.spec.ts
8 passed (16.5s)
```

- The deterministic-marker test (`_fazMaybeSwapLanguage does not fetch when navigator language matches server default`) now completes in 293ms vs the previous 750ms fixed sleep — no more racing the REST layer.
- The `multilingual-cache.spec.ts:141` test (`navigator.languages override triggers a client-side swap`) which was flaky under full-suite load also passes cleanly (831ms).

## Intentionally out of scope

A grep over the rest of the spec directory found 15 other files with top-level `const WP_BASE = process.env.WP_BASE_URL ?? 'http://localhost:9998'` declarations. They don't share the `beforeAll` pain point that motivated this PR (the top-level const is evaluated once per module and doesn't drift), and migrating them all at once would turn a targeted fixture improvement into a 15-file refactor. Separate follow-up PR recommended if the team wants every spec to consume the fixture.

Also unchanged: reCAPTCHA providers remain classified as `necessary` per user decision on #69 — the JSON schema validation added there stays in place.

## CodeRabbit

Triggering `@coderabbitai review` below to run the full review; previous rate limit on #69 should not carry over to this PR.


fabiodalez-dev:
@coderabbitai full review

PR #91, opened 2026-04-27
fix(#87): inject video placeholder even when iframe has zero offsetWidth/Height

### LINKED ISSUE #87: Video (Youtube) element is blocked, but no consent placeholder is shown
Hello,
I built a page using the Bricks Builder and used the native Video element to embed a YouTube video. The cookie banner correctly detects and blocks the script, and I also added “YouTube” manually from the blocking templates to be safe.

However, instead of the video element being replaced with a consent placeholder/banner, nothing is displayed. In the page source I can see that something should be rendered there, but it is not visible on the frontend.

So the YouTube embed is blocked, but the user does not get a way to give consent for this specific element:
Image
Image

fabiodalez-dev:
## Issue

[#87](https://github.com/fabiodalez-dev/FAZ-Cookie-Manager/issues/87) — Bricks Builder Video element is blocked but no consent placeholder appears, just an empty wrapper.

## Root cause (the real one this time)

`_fazAddPlaceholder()` in `frontend/js/script.js` had an early return:

```js
const { offsetWidth, offsetHeight } = htmlElm;
if (offsetWidth === 0 || offsetHeight === 0) return;
```

Page-builder Video elements stage the iframe inside a wrapper that uses `aspect-ratio: 16/9` with no explicit width/height on the iframe itself. At MutationObserver time the iframe's own `offsetWidth`/`offsetHeight` are still 0 — the browser hasn't run layout yet — so the early-return bailed out and the placeholder was *never injected*. End result: empty Bricks wrapper, no consent CTA.

The 1.13.7 fix only handled the *adjacent* CSS-collapse case (placeholder injected but collapsing because of `min-height: 0`). It didn't address the never-injected case 3DRZ actually reported.

## Fix — 4-step fallback chain

Replace the silent bail-out with:

1. The iframe's own `offsetWidth`/`offsetHeight` (the original behaviour).
2. **Walk up to 4 ancestors** looking for a measured box. The Bricks `<div class="bricks-video">` wrapper, the Elementor `<div class="elementor-video-wrapper">` wrapper, and Divi's `<div class="et_pb_video">` wrapper all carry a real layout box even when the inner iframe doesn't.
3. **`requestAnimationFrame` retry**, up to 3 frames, so the browser has time to settle layout if the iframe was just inserted.
4. **Last resort** — inject the placeholder anyway with no inline `width`/`height` so the existing CSS floor takes over (`min-height: 200px` on `.faz-placeholder`, `aspect-ratio: 16/9` on `.faz-placeholder--video`). The visitor always sees the Cookie-required CTA.

Also hardened the surrounding flow:
- Optional-chain the shortcode lookup (don't crash if the bundle ships without `faz_video_placeholder`).
- Guard the `parentNode` check (the iframe may have been detached between observer fire and our turn).
- Tolerate a null `innerTextElement` if the placeholder template was overridden by a theme.

## Test plan

- [ ] Manual: Bricks 1.x install on WP 6.9 + Divi/Elementor sanity passes.
- [ ] Repro from #87 (the 3DRZ Bricks Video case): no more empty wrapper, placeholder injected with the CTA visible at 16:9.
- [ ] Existing E2E suite continues to pass (the fix is strictly additive — never-injected → injected; previously-injected continues unchanged).

## Note on lightbox case (separate)

3DRZ also reported a Bricks lightbox-link case (`<a>` with a YouTube URL that opens in a Bricks lightbox). That is a *different* code path — JS-driven popup that the scanner can't see until the user clicks. It needs a click-time interceptor and ships separately when we have a public repro URL.

Closes #87.

## Summary by CodeRabbit

* **Bug Fixes**
  * Migliorata gestione dei placeholder video quando le dimensioni non sono disponibili, evitando aborti e garantendo che i segnaposto vengano inseriti e ridimensionati correttamente.
  * Resilienza nell'estrazione degli ID YouTube quando gli attributi mancanti.

* **New Features**
  * Blocco lato client dell'apertura di lightbox/modali di page-builder quando il consenso impedisce il caricamento del provider video, con inserimento di un segnaposto UX.

* **Tests**
  * Aggiunti test end-to-end per riprodurre e verificare i comportamenti sopra descritti, incluse verifiche in contesti Bricks.


fabiodalez-dev:
@coderabbitai thanks — the critical finding is correct and now fixed in `e87cb9b`.

The race you identified was real: `_fazMutationObserver` calls `_fazAddPlaceholder()` on line ~1707 and then issues `node.remove()` on line ~1752 within the same synchronous tick. With the previous implementation deferring the insert to `requestAnimationFrame`, the iframe was detached from the DOM before the rAF fired, the `parentNode` guard tripped, and the placeholder was silently dropped — exactly the empty-wrapper regression the PR was supposed to fix.

Fix in `e87cb9b` flips the contract:
- The placeholder is **always inserted synchronously** on the first call, before the calling site can issue `node.remove()`. If `width`/`height` are still 0 after the ancestor walk, we omit inline sizing and let the CSS floor (`min-height: 200px`, `aspect-ratio: 16/9` on `.faz-placeholder--video`) own the layout. Visitor always sees the CTA.
- The `requestAnimationFrame` is preserved but its scope is now strictly **remeasure-and-tighten**: at the next frame the iframe is gone, so we re-walk the placeholder's own ancestor chain and update `style.width` only if a non-zero measurement materialises and we hadn't already pinned it. The rAF only ever mutates the existing placeholder; it never inserts.

Both tests on the branch still pass:
- `bricks-iframe-repro.mjs` — placeholder count 0 → 1
- `bricks-real-post.mjs` — placeholder injected, YouTube iframe NOT executed, console errors = 0 (verified against Bricks 2.3.4 active as the theme)

---

Side note on the nitpick about redundant `comment_author_email_` / `comment_author_url_` prefixes — that one was already addressed in PR #89 (commit `dd4f433`, merged into `main`), and the current branch sits on top of that merge. The `frontend/class-frontend.php` content on `fix/87-bricks-iframe-placeholder` already has just the single `'comment_author_'` entry plus a comment documenting the three-variant coverage. The walkthrough must have been generated against an out-of-date base; nothing to do here.

Issue #99, opened 2026-05-15
Bug: wp_localize_script payload (`{handle}-js-extra`) can be blocked as analytics — breaks bootstrap (`_fazCfg` / `_fazConfig` undefined before `faz-fw-js-after`)


Myblueroom:
### Summary
Under “Alternative asset path” (`faz-fw` handle), the localized store from `wp_localize_script()` is printed as `<script id="faz-fw-js-extra">` with `var _fazCfg = { … }`.

The script-blocking pipeline (`filter_inline_script_tag` and/or the output-buffer pass in `process_script_tag`) can rewrite **that same** inline tag to `type="text/plain"` and add `data-faz-category="…"` when a provider scanner pattern matches **inside the JSON body** (cookie catalog / vendor strings / URLs are enough for false positives).

Meanwhile `faz-fw-js-before` / `faz-fw-js-after` remain executable `text/javascript`, including the bridge `window._fazConfig = window._fazCfg` and the main bundle that starts with `const _fazStore = window._fazConfig`.

Result: `_fazCfg` never executes → `_fazConfig` is undefined → `TypeError` on `_backupNodes` (or equivalent) and no `#faz-consent` / broken CMP.

This is a chicken-and-egg: the CMP cannot load because its own config script is classified as blocked analytics.

### Environment
- FAZ Cookie Manager: **1.13.16**
- WordPress: **updated version**
- “Alternative asset path”: **enabled**

### Evidence (view-source)
- `id="faz-fw-js-extra"`: was `type="text/plain"` + `data-faz-category="analytics"` while containing `var _fazCfg = …`
- `id="faz-fw-js-before"` / `id="faz-fw-js-after"`: `type="text/javascript"` and executed immediately

### Expected
The bootstrap localized payload for FAZ’s own enqueue handles must **never** be neutralised by the scanner/blocker, or it must always execute before the main bootstrap bundle.

### Suggested fix (what worked for us)
1. In `filter_inline_script_tag()`: if `$id` ends with `-js-extra` and the derived handle is an “own” FAZ handle (`is_own_script_handle()`), return `$tag` unchanged.
2. In `process_script_tag()` (output buffer): same exemption for `<script id="…-js-extra">` owned by FAZ, so the OB pass cannot re-block it.

Additionally, harden the alt-asset bridge to avoid assigning `undefined`:
[...]

PR #101, opened 2026-05-15
Fix: do not block FAZ own *-js-extra localized payload (false positives)

Myblueroom:
Summary
With “Alternative asset path” enabled (faz-fw handle), the localized store from wp_localize_script() is printed as <script id="faz-fw-js-extra"> containing var _fazCfg = { … }.

The script-blocking pipeline (filter_inline_script_tag and/or the output-buffer pass in process_script_tag) can rewrite that same inline tag to type="text/plain" and add data-faz-category="…" when a provider scanner pattern matches inside the JSON body (cookie catalog / vendor strings / URLs are enough for false positives).

Meanwhile faz-fw-js-before / faz-fw-js-after remain executable text/javascript, including the bridge window._fazConfig = window._fazCfg and the main bundle that starts with const _fazStore = window._fazConfig.

Result: _fazCfg never executes → _fazConfig is undefined → TypeError on _backupNodes (or equivalent) and a broken CMP / no #faz-consent.

This is a chicken-and-egg: the CMP cannot load because its own localized config script is classified as blocked analytics.

Environment
FAZ Cookie Manager: 1.13.16
WordPress: 6.9.4
PHP: 8.1
Theme: Bimber
“Alternative asset path”: enabled
Evidence (view-source)
id="faz-fw-js-extra": type="text/plain" + data-faz-category="analytics" while containing var _fazCfg = …
id="faz-fw-js-before" / id="faz-fw-js-after": type="text/javascript" and executed immediately
Expected
The bootstrap localized payload for FAZ’s own enqueue handles must never be neutralised by the scanner/blocker, or it must always execute before the main bootstrap bundle.

Suggested fix (worked locally)
In filter_inline_script_tag(): if $id ends with -js-extra and the derived handle passes is_own_script_handle(), return $tag unchanged.
In process_script_tag() (output buffer): same exemption for <script id="…-js-extra"> owned by FAZ.
Additionally harden the alt-asset bridge to avoid assigning undefined:
if (typeof window._fazCfg !== 'undefined' && window._fazCfg !== null) { window._fazConfig = window._fazCfg; }

Optional in script.js:
window._fazConfig = window._fazConfig ?? window._fazCfg;

Note on repro / routing
On some sites, “Script blocking → Excluded pages” can disable blocking on certain paths, masking the bug on one URL while the homepage still breaks.

Console (representative)
Uncaught TypeError: Cannot set properties of undefined (setting '_backupNodes') at faz-fw-js-after:…

Thanks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Risolto problema di inizializzazione della configurazione in scenari di ordine di esecuzione variabile (cache e guest-mode), garantendo il corretto caricamento dei parametri di configurazione indipendentemente dal timing di esecuzione.
  * Migliorata gestione degli script inline per evitare neutralizzazione errata dei payload di configurazione del plugin.

[Review Change Stack](https://app.coderabbit.ai/change-stack/fabiodalez-dev/FAZ-Cookie-Manager/pull/101)


fabiodalez-dev:
Thank you @Myblueroom for reporting this independently — knowing that multiple users hit the same wall was a useful confirmation that the fix needed to ship in 1.13.17.

The underlying bug was already resolved in v1.13.17 (commit 57293b8, closes #99), which is essentially equivalent to your patch: `is_own_script_handle()` / `is_own_inline_script_id()` helpers in `frontend/class-frontend.php` guard both `filter_inline_script_tag` and the output-buffer pass against rewriting our own `-js-extra` payloads. The implementation differs in style (helper functions vs. inline conditionals in `process_script_tag` / `filter_inline_script_tag`), but the behavior is equivalent.

Closing as already resolved. You're credited in `CHANGELOG.md`, `readme.txt`, the GitHub release notes, and `README.md` for the independent confirmation.

— Release notes: https://github.com/fabiodalez-dev/FAZ-Cookie-Manager/releases/tag/v1.13.17

Issue #103, opened 2026-05-15
Request new features and Improvement


Zeequeit:
Hello i appreciate your efforts and hard work you have put in this CMP, i want to request a function to add more than one Cookie banner which we can used separately in the US using Geo based location, which only show the opt out option in the US as US doesn't require the visitors to directly block the cookies or getting data, it just require the sites to show an option for opt out from collecting the cookies, unless a user specifically doesn't  ask to opt out, blocking the cookies are not necessary, but if we use the same cookie banner of EEA/EU countries in the US, that really limit the original  full permissions granted in US, and also the banner can be a lot smaller with just one opt out option for the US visitors, so by adding to option to create another banner or multiple banners which show based on separate geo targeting condition is very useful feature. 

My second request is to apply for the google CMP certification, as a lot of users use the Consent banners just for the sake of Google AdSense ads and increasing the ad revenue on their site, if this plugin get certified by the Google, it would be best and it can also lead a lot of users to the plugin as it would be listed in the google certified CMP partner list. since Google only accept the CMP signals for their ads which are certified by the Google directly. I hope I would get a positive response on my request.

<img width="773" height="549" alt="Image" src="https://github.com/user-attachments/assets/e0218feb-1423-47c4-8d25-f3647874288f" />

Zeequeit:
<img width="617" height="734" alt="Image" src="https://github.com/user-attachments/assets/88010c5b-9990-458b-866e-cc3f814b56d5" />

fabiodalez-dev:
Hi @Zeequeit, let me reply to both requests separately, because they sit in very different places.

Quick context before the two answers: FAZ Cookie Manager is fully free, no paid tier, no cloud account, no telemetry — I do not receive any compensation from this project. I build and maintain it in my spare time, alone. That shapes both answers below.

**On the multi-banner / geo-routing request (US opt-out only, EU full GDPR):**

Your reasoning is correct from a compliance standpoint. CCPA / CPRA does not require the same opt-in-then-block model that GDPR/ePrivacy does, so serving the full EU banner to US visitors is unnecessarily restrictive (and as you said, hurts the AdSense revenue side without legal justification).

[...]

PR #104, opened 2026-05-16
feat(banners): multi-banner geo-routing (US opt-out vs EU GDPR, refs #103)

fabiodalez-dev:
## Summary

Closes the routing gap raised in #103: serve a CCPA-light banner to US visitors and a full GDPR banner to EU visitors out of the same install, picked automatically by the visitor's detected country.

The plugin had the foundation already (multi-row \`wp_faz_banners\` table, GeoIP detection, region map covering EU/UK/US/CA/BR/AU/JP/CH), but the runtime picked the first \`status=1\` banner regardless of geo. This PR adds the routing layer.

### DB schema (\`wp_faz_banners\`)
- New \`target_countries longtext\` — JSON array of ISO-3166 alpha-2 codes (e.g. \`[\"US\",\"CA\"]\`). Empty array \`[]\` = match every visitor (preserves single-banner installs).
- New \`priority int\` — tie-break when multiple banners target the same country. Higher wins.
- Idempotent migration \`update_db_350()\` re-runs \`dbDelta\` for column-add, backfills existing rows to \`target_countries='[]'\`, promotes the first \`status=1\` banner to \`banner_default=1\` if none was flagged.

### Banner model (\`Banner\`)
- \`set_target_countries()\` normalises input (case, whitespace, dedup, drops non-2-letter tokens).
- \`set_priority()\` clamps negatives to 0.
- Matching getters expose the persisted values.

### Selector (\`Controller::get_active_banner_for_country(\$country)\`)
Picks in this order:

1. \`status=1\` banner whose \`target_countries\` contains \`\$country\` (priority desc, then \`banner_id\` asc for deterministic ties).
2. \`status=1\` banner with empty \`target_countries\` (match-all rows).
3. \`banner_default=1\` row (any status) — last-resort fallback.

\`get_active_banner()\` now delegates to \`get_active_banner_for_country('')\` for full backward compatibility.

### Frontend wiring (\`class-frontend.php\`)
- New private \`get_visitor_country()\` — same CF-IPCountry + MaxMind chain as \`is_geo_banner_disabled()\`.
- New \`faz_visitor_country\` filter lets test fixtures and edge deployments stub the country.
- \`enqueue_scripts()\` feeds the detected country into the new selector.

### Admin UI (Banner edit page)
New **Geo Targeting** tab:
- Region preset toggles (EU/EEA, UK, US, CA, BR, AU, JP, CH).
- Custom country list (comma-separated codes for NZ/SG/KR/etc.).
- Priority integer input.
- \"Use this banner as the default fallback\" toggle (\`banner_default\` flag).

JS does round-trip-lossless serialisation: reverses persisted \`target_countries\` back into region presets + leftover custom codes on load, merges + normalises on save.

### REST API
- \`get_item_schema()\` declares \`target_countries\` (array of \`/^[A-Z]{2}\$/\` strings) and \`priority\` (non-negative int).
- \`prepare_item_for_database()\` reads both fields when present in the request (optional — legacy clients keep working).
- \`get_formatted_item_data()\` emits both fields in the response.

## Test plan

- [x] **GEO-01**: country=US → US-targeted banner; country=IT → match-all fallback
- [x] **GEO-02**: priority breaks ties (higher wins)
- [x] **GEO-03**: \`banner_default=1\` is the last-resort fallback when nothing else matches
- [x] **GEO-04**: \`set_target_countries\` normalisation (case/whitespace/dedup/invalid drop)
- [x] **GEO-05**: \`set_priority\` clamps negatives to 0
- [x] **GEO-06**: REST \`GET /banners/{id}?context=edit\` round-trips the new fields
- [x] **GEO-07**: \`faz_visitor_country\` filter steers the picker

All 7 pass in 10.9s on the local nginx+PHP-FPM stack.

## Migration safety

- Single-banner installs (the 99% baseline): \`target_countries='[]'\` after migration → still match every visitor → no behaviour change.
- One row is always promoted to \`banner_default=1\` so the last-resort fallback is never empty.
- Legacy callers of \`get_active_banner()\` still work (delegates to \`get_active_banner_for_country('')\`).
- REST clients that don't send the new fields keep working (optional in \`prepare_item_for_database\`).

## Out of scope (future)

- IPv4/IPv6 detection beyond MaxMind + CF-IPCountry + ip-api.com (the three sources the plugin already uses).
- Per-page geo overrides (the existing \`excluded_pages\` setting handles \"hide banner here\" — orthogonal to which banner is chosen when shown).
- WP-CLI commands for bulk banner-geo configuration (admin UI is the canonical entry point).

## Version

Bumped to **1.14.0** (minor feature release). The 1.13.18 patch on #102 lives in a separate branch and can merge in either order.

Refs #103.

## Summary by CodeRabbit

* **New Features**
  * Geo Targeting: nuova scheda per regioni/paesi (preset o custom), priorità e flag di fallback; multi-banner selezionabile per paese; esposizione/persistenza di target_countries e priority via API, front-end e AMP.

* **UX**
  * Switcher per navigare tra banner, modal “+ New banner”, rename/delete inline; toggle per-banner “Allow Close (X) alongside Reject”.

* **Bug Fixes / Reliability**
  * Unico default garantito, recovery UI per banner mancanti, cache/header adattati per output dipendente dal paese.

* **Tests / Chores**
  * Ampia batteria e2e aggiornata, DB migration e versione 1.14.0, documentazione ampliata.

[Review Change Stack](https://app.coderabbit.ai/change-stack/fabiodalez-dev/FAZ-Cookie-Manager/pull/104?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai full review

Latest batch of fixes since the previous review (commits 4e89a61 → 87d1bb9):

1. **fix(create_item id pollution)** — `Controller::create_item` no longer re-reads `$wpdb->insert_id` after `$banner->save()` (downstream INSERTs polluted it with wp_options.AUTO_INCREMENT, producing phantom banner ids in admin redirect URLs).
2. **fix(REST get_item 404)** — `GET /faz/v1/banners/{id}` now does an explicit existence probe and returns a structured 404 + `fazcookie_rest_invalid_id` for non-existent ids instead of a 200 with empty payload.
3. **fix(REST delete_item 404)** — same existence-probe pattern for DELETE. JS handler now treats 404 as 'already removed → reload to default' instead of a hard error.
4. **fix(missing-banner notice UI)** — `admin/views/banner.php` + `banner.js` show an in-page recovery card with a working CTA pointing to the actual `banner_default=1` row.
5. **feat(UX consolidation)** — toolbar dropdown replaced with always-visible chip buttons; inline rename moved from toolbar to a 'Banner Name' card in the General tab; Geo Targeting promoted from 7th to 2nd tab position.
6. **fix(admin no-cache)** — `Admin::render_page` now stacks every available no-cache primitive (nocache_headers, Cache-Control no-store, X-LiteSpeed-Cache-Control, DONOTCACHE* constants, litespeed_control_set_nocache action).
7. **fix(REST no-cache)** — global `rest_pre_dispatch` hook on `/faz/v1/*` fires `litespeed_control_set_nocache` + emits `X-LiteSpeed-Cache-Control: no-cache` BEFORE the route runs (response-time `Cache-Control: no-store` wasn't enough — LSCache's private-cache mode ignores it on routes the site has opted into).
8. **fix(cache module load order)** — `Cache::init()` now runs `load_services()` immediately when `plugins_loaded` has already fired (REST/admin contexts). Pre-fix, `Litespeed_Cache::clear_cache` and every other cache-adapter never registered on `faz_after_update_banner` → no third-party cache was ever purged on banner CRUD.

New regression tests: `pr104-create-id-pollution`, `pr104-missing-banner-notice`, `pr104-delete-stale-id`, `pr104-admin-nocache`, `pr104-rest-nocache`, `pr104-issue-fixes` (covering issues #108-#113).

Please review the new commits with particular attention to:
- The `rest_pre_dispatch` filter signature and `did_action(plugins_loaded)` guard for any edge cases I missed
- The DB existence probe pattern duplicated in get_item / delete_item — possibly worth extracting into a Controller helper
- The redirect-on-404 UX flow in banner.js — is the 700ms setTimeout enough on slow networks?

fabiodalez-dev:
> ✅ **17 of 20 findings fixed** (commits `5e84413` + `2fb77e5`).
> Live verified on prod (fabiodalez.it):
> - `/faz/v1/banner/{lang}` now `x-litespeed-cache-control: public,max-age=604800` (F004 ✓)
> - `/faz/v1/banners` still `no-cache` (admin REST, F004 boundary preserved ✓)
> - create → delete cycle: 200/200, second delete 404 (F016/F017 atomic transaction ✓)
> - chip switcher renders correctly post-double-fire-removal (F006/F007 ✓)
> 
> **Resolved (17):** F001 F002 F003 F004 F005 F006 F007 F008 F010 F011 F012 F013 F015 F016 F017 F019 F020
> **Skipped (3, audit-only):** F009 (dup of F002), F014 (cosmetic), F018 (design discussion)
> 
> Smoke: 41/41 PR104 + multi-banner-geo-routing specs green in 35.6s post-fix.


> ✅ **Fixes applied (commit `5e84413`).** 13 of 20 findings landed in `feat/multi-banner-geo-routing`:
> - **Critical regressions** (this PR's own commits): F001, F004, F006, F007, F011, F015, F016
> - **Privacy / security**: F019, F020
> - **Other bugs**: F002, F003, F005, F008, F012, F013
> 
> **Skipped** (audit trail in artifact): F009 (dup of F002), F010 (refactor punted), F014 (cosmetic), F017 (single-admin), F018 (design discussion).
> 
> Smoke: 11/11 PR104 specs green post-fix.


> ⚠️ **Reduced-scope review.** L3 (CLAUDE.md), L4 (comments), L5 (UX), L7 (holistic) skipped to keep cost bounded on a 45-commit / 13k-line PR.  
> Lanes run: **L1** (diff-local, Sonnet) + **L2** (structural/blast-radius, Opus) + **L6** (security, Sonnet).  
> Phase 4 validation was inline-scored — every finding is `pending_validation` upgraded to `confirmed_mechanical` or `partial` based on lens origin_confidence × impact_type only. **Treat the deep-lane findings as starting points, not verdicts.**


### Code review

**Branch:** `feat/multi-banner-geo-routing` → `main`
**PR:** #104 (open)
**Review ID:** `rev_01KRXMJG6JPJ4ZVYJYB703SF6M`
**Fix threshold:** not yet set (run `/adamsreview:fix [threshold]` to apply fixes)
**Sub-agent tokens:** 0 across 0 invocations

Found 20 findings across all lanes:
- Deep lane (correctness/security): 11 auto-fixable, 9 partial

---

### ✓ Auto-fixable (20)

| F001 | 80 | correctness | `admin/assets/js/pages/banner.js:983-987` | bindCloseSubToggle reads .checked on a <label> element. getElementById('faz-b-close-toggle') returns the <label> wrapper (confirmed in admin/views/banner.php line 559), not the <input type='checkbox'> inside it. label.checked is undefined; !!undefined === false, so the sub-toggle is unconditionally disabled regardless of the parent toggle state. |
| F002 | 80 | correctness | `includes/class-activator.php:784` | Migration update_db_350() adds target_countries as NULL-able (longtext NULL) but the canonical CREATE TABLE schema in class-controller.php declares target_countries longtext NOT NULL. On a fresh install the column is NOT NULL; on an upgrade it is NULL-able. |
| F003 | 70 | correctness | `admin/modules/banners/api/class-api.php:287-295` | delete_item() validates $banner_id <= 0 and returns a WP_Error with code 'fazcookie_rest_item_exists'. The code name is semantically inverted — 'item_exists' is the wrong code for an invalid/missing ID guard. |
| F004 | 85 | correctness | `faz-cookie-manager.php:485-502` | Newly-added rest_pre_dispatch filter sets X-LiteSpeed-Cache-Control: no-cache and fires litespeed_control_set_nocache for EVERY /faz/v1/* REST route, including the public /faz/v1/banner/{lang} endpoint which the banner-rest callback explicitly tries to cache with 'Cache-Control: public, max-age=300'. Missing the '/faz/v1/banner' exclusion that the sister filter in admin/class-admin.php::add_rest_nocache_headers() carries, this defeats the intentional 5-minute CDN caching of the public banner payload. |
| F005 | 85 | correctness | `frontend/js/script.js:201-229` | _fazReadScopedCookieValue() falls back to the unprefixed 'banner'/'law' cookie keys for backwards compat, but category slugs are sanitize_title()'d which permits 'banner' or 'law' as legitimate category slugs. A returning visitor whose install has a category slug exactly equal to 'banner' or 'law' will see storedBannerSlug='yes' (the category's consent value) compared against the actual banner slug like 'gdpr-1' — _fazConsentScopeChanged returns true, the consent cookie is invalidated and the banner re-pops on every single page load. |
| F006 | 85 | correctness | `admin/modules/banners/includes/class-controller.php:197-220` | create_item() unconditionally calls do_action('faz_after_update_banner') and delete_cache() at the end, but $banner->save() at the previous line internally routes through update_item() — which itself fires do_action('faz_after_update_banner') + delete_cache(). Every banner CREATE fires faz_after_update_banner twice and runs delete_cache() at least three times back-to-back. Cache adapters purge LSCache/Rocket twice per create. |
| F007 | 85 | correctness | `admin/modules/banners/includes/class-controller.php:196-220` | create_item() calls clear_default_on_others($id) right after the INSERT, but then save() routes to update_item() which reads was_default=true and runs clear_default_on_others($id) a SECOND time. The epoch bumps four times for a single CREATE. |
| F008 | 75 | correctness | `frontend/class-frontend.php:936-945` | Frontend::is_country_in_regions() defines the 'eu' region preset to INCLUDE GB, but the admin view in admin/views/banner.php explicitly says 'EU preset deliberately excludes GB; UK is its own preset' and admin/assets/js/pages/banner.js REGION_PRESETS.EU excludes GB. Two parallel preset definitions for the same concept have diverged. |
| F009 | 75 | correctness | `includes/class-activator.php:749-800` | update_db_350() safety-net at line 784 re-adds target_countries as 'longtext NULL' — the opposite nullability of the schema NOT NULL. The DB column ends up nullable; subsequent dbDelta passes attempt to ALTER COLUMN NOT NULL and the schema-vs-actual drift persists across versions. |
| F010 | 75 | correctness | `admin/modules/banners/includes/class-controller.php:419-486` | get_active_banner_for_country() instantiates 'new Banner( $item->banner_id )' on every iteration even though prepare_item() already loaded the row. Each Banner() constructor invokes a Banner-controller round-trip + full sanitize_settings cascade. On installs with N banners this turns a single get_active_banner call into N redundant DB reads. |
| F011 | 75 | correctness | `admin/modules/banners/includes/class-controller.php:518-579` | has_country_dependent_banners() uses get_option('faz_banner_cache_epoch') with autoload=false set by delete_cache()'s update_option call. With autoload=false the alloptions cache no longer carries this option, defeating the cross-node propagation path the comment promises. |
| F012 | 85 | correctness | `includes/class-i18n-helpers.php:275-320` | faz_country_to_locale()'s map contains the synthetic key 'CA_FR' => 'fr_CA' but get_visitor_country() returns 2-letter ISO-3166 alpha-2 codes only. The 5-character key is unreachable; no real-world country lookup will resolve to 'fr_CA'. |
| F013 | 75 | correctness | `frontend/class-amp-consent.php:340-360` | AMP_Consent::get_active_banner() resolves the banner from the visitor's country but the AMP class never emits DONOTCACHEPAGE / Cache-Control: no-store headers when has_country_dependent_banners() is true. A page-cache plugin storing the AMP HTML for an EU visitor will serve the same EU-banner AMP page to a non-EU visitor. |
| F014 | 65 | correctness | `admin/modules/banners/api/class-api.php:102-110` | The 'id' sanitize_callback=absint is on the OUTER 'args' block, but the INNER per-method 'args' blocks for create/update/delete each declare their own 'args' override that REPLACES the outer dict in WP_REST_Server's arg merging. Net effect: absint sanitize is honoured only for GET; silently dropped for DELETE. |
| F015 | 75 | correctness | `admin/modules/banners/includes/class-controller.php:596-606` | delete_cache() computes $next = $current + 1; two concurrent admin delete_cache() calls compute the same $next, write the same new key, and one writer may overwrite the other's fresher value under the same cache key for the full 5-minute TTL. |
| F016 | 75 | correctness | `admin/modules/banners/includes/class-controller.php:298-324` | delete_item() captures $was_default via SELECT then $wpdb->delete then promote_fallback_default — non-atomic. Two concurrent admin sessions deleting two different banners can each promote a peer, ending with TWO default rows. The at-most-one invariant is violated. |
| F017 | 85 | security | `admin/modules/banners/api/class-api.php:324-331` | TOCTOU race between banner_exists() SELECT probe and controller->delete_item() DELETE: two concurrent admins can both pass existence check before either delete fires. |
| F018 | 85 | security | `frontend/class-frontend.php:1082` | _scopeFingerprint embeds a 32-hex substring of wp_hash(slug|law, 'auth') in every public page HTML: fingerprint keyed by site's auth secret (wp_salt('auth')); repeated oracle access with known inputs could support offline brute-force against weak salts. |
| F019 | 85 | security | `includes/class-geolocation.php:218-222` | Raw visitor IP passed as third argument to faz_country_detection_consensus filter; third-party plugins receive unhashed PII without notice or hashing. |
| F020 | 85 | security | `uninstall.php:286-312` | Multisite sitemeta transient sweep runs unconditionally inside is_multisite() with no faz_should_remove_on_uninstall() gate: every network-level FAZ transient deleted on any multisite uninstall regardless of admin choice, violating remove_data_on_uninstall=false contract. |

#### F001 — bindCloseSubToggle reads .checked on a <label> element. getElementById('faz-b-close-toggle') returns the <label> wrapper (confirmed in admin/views/banner.php line 559), not the <input type='checkbox'> inside it. label.checked is undefined; !!undefined === false, so the sub-toggle is unconditionally disabled regardless of the parent toggle state.

**File:** `admin/assets/js/pages/banner.js:983-987`
**Score:** 80 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F002 — Migration update_db_350() adds target_countries as NULL-able (longtext NULL) but the canonical CREATE TABLE schema in class-controller.php declares target_countries longtext NOT NULL. On a fresh install the column is NOT NULL; on an upgrade it is NULL-able.

**File:** `includes/class-activator.php:784`
**Score:** 80 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F003 — delete_item() validates $banner_id <= 0 and returns a WP_Error with code 'fazcookie_rest_item_exists'. The code name is semantically inverted — 'item_exists' is the wrong code for an invalid/missing ID guard.

**File:** `admin/modules/banners/api/class-api.php:287-295`
**Score:** 70 (moderate)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F004 — Newly-added rest_pre_dispatch filter sets X-LiteSpeed-Cache-Control: no-cache and fires litespeed_control_set_nocache for EVERY /faz/v1/* REST route, including the public /faz/v1/banner/{lang} endpoint which the banner-rest callback explicitly tries to cache with 'Cache-Control: public, max-age=300'. Missing the '/faz/v1/banner' exclusion that the sister filter in admin/class-admin.php::add_rest_nocache_headers() carries, this defeats the intentional 5-minute CDN caching of the public banner payload.

**File:** `faz-cookie-manager.php:485-502`
**Score:** 85 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F005 — _fazReadScopedCookieValue() falls back to the unprefixed 'banner'/'law' cookie keys for backwards compat, but category slugs are sanitize_title()'d which permits 'banner' or 'law' as legitimate category slugs. A returning visitor whose install has a category slug exactly equal to 'banner' or 'law' will see storedBannerSlug='yes' (the category's consent value) compared against the actual banner slug like 'gdpr-1' — _fazConsentScopeChanged returns true, the consent cookie is invalidated and the banner re-pops on every single page load.

**File:** `frontend/js/script.js:201-229`
**Score:** 85 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F006 — create_item() unconditionally calls do_action('faz_after_update_banner') and delete_cache() at the end, but $banner->save() at the previous line internally routes through update_item() — which itself fires do_action('faz_after_update_banner') + delete_cache(). Every banner CREATE fires faz_after_update_banner twice and runs delete_cache() at least three times back-to-back. Cache adapters purge LSCache/Rocket twice per create.

**File:** `admin/modules/banners/includes/class-controller.php:197-220`
**Score:** 85 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F007 — create_item() calls clear_default_on_others($id) right after the INSERT, but then save() routes to update_item() which reads was_default=true and runs clear_default_on_others($id) a SECOND time. The epoch bumps four times for a single CREATE.

**File:** `admin/modules/banners/includes/class-controller.php:196-220`
**Score:** 85 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F008 — Frontend::is_country_in_regions() defines the 'eu' region preset to INCLUDE GB, but the admin view in admin/views/banner.php explicitly says 'EU preset deliberately excludes GB; UK is its own preset' and admin/assets/js/pages/banner.js REGION_PRESETS.EU excludes GB. Two parallel preset definitions for the same concept have diverged.

**File:** `frontend/class-frontend.php:936-945`
**Score:** 75 (moderate)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F009 — update_db_350() safety-net at line 784 re-adds target_countries as 'longtext NULL' — the opposite nullability of the schema NOT NULL. The DB column ends up nullable; subsequent dbDelta passes attempt to ALTER COLUMN NOT NULL and the schema-vs-actual drift persists across versions.

**File:** `includes/class-activator.php:749-800`
**Score:** 75 (moderate)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F010 — get_active_banner_for_country() instantiates 'new Banner( $item->banner_id )' on every iteration even though prepare_item() already loaded the row. Each Banner() constructor invokes a Banner-controller round-trip + full sanitize_settings cascade. On installs with N banners this turns a single get_active_banner call into N redundant DB reads.

**File:** `admin/modules/banners/includes/class-controller.php:419-486`
**Score:** 75 (moderate)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F011 — has_country_dependent_banners() uses get_option('faz_banner_cache_epoch') with autoload=false set by delete_cache()'s update_option call. With autoload=false the alloptions cache no longer carries this option, defeating the cross-node propagation path the comment promises.

**File:** `admin/modules/banners/includes/class-controller.php:518-579`
**Score:** 75 (moderate)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F012 — faz_country_to_locale()'s map contains the synthetic key 'CA_FR' => 'fr_CA' but get_visitor_country() returns 2-letter ISO-3166 alpha-2 codes only. The 5-character key is unreachable; no real-world country lookup will resolve to 'fr_CA'.

**File:** `includes/class-i18n-helpers.php:275-320`
**Score:** 85 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F013 — AMP_Consent::get_active_banner() resolves the banner from the visitor's country but the AMP class never emits DONOTCACHEPAGE / Cache-Control: no-store headers when has_country_dependent_banners() is true. A page-cache plugin storing the AMP HTML for an EU visitor will serve the same EU-banner AMP page to a non-EU visitor.

**File:** `frontend/class-amp-consent.php:340-360`
**Score:** 75 (moderate)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F014 — The 'id' sanitize_callback=absint is on the OUTER 'args' block, but the INNER per-method 'args' blocks for create/update/delete each declare their own 'args' override that REPLACES the outer dict in WP_REST_Server's arg merging. Net effect: absint sanitize is honoured only for GET; silently dropped for DELETE.

**File:** `admin/modules/banners/api/class-api.php:102-110`
**Score:** 65 (moderate)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F015 — delete_cache() computes $next = $current + 1; two concurrent admin delete_cache() calls compute the same $next, write the same new key, and one writer may overwrite the other's fresher value under the same cache key for the full 5-minute TTL.

**File:** `admin/modules/banners/includes/class-controller.php:596-606`
**Score:** 75 (moderate)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F016 — delete_item() captures $was_default via SELECT then $wpdb->delete then promote_fallback_default — non-atomic. Two concurrent admin sessions deleting two different banners can each promote a peer, ending with TWO default rows. The at-most-one invariant is violated.

**File:** `admin/modules/banners/includes/class-controller.php:298-324`
**Score:** 75 (moderate)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F017 — TOCTOU race between banner_exists() SELECT probe and controller->delete_item() DELETE: two concurrent admins can both pass existence check before either delete fires.

**File:** `admin/modules/banners/api/class-api.php:324-331`
**Score:** 85 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F018 — _scopeFingerprint embeds a 32-hex substring of wp_hash(slug|law, 'auth') in every public page HTML: fingerprint keyed by site's auth secret (wp_salt('auth')); repeated oracle access with known inputs could support offline brute-force against weak salts.

**File:** `frontend/class-frontend.php:1082`
**Score:** 85 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F019 — Raw visitor IP passed as third argument to faz_country_detection_consensus filter; third-party plugins receive unhashed PII without notice or hashing.

**File:** `includes/class-geolocation.php:218-222`
**Score:** 85 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

#### F020 — Multisite sitemeta transient sweep runs unconditionally inside is_multisite() with no faz_should_remove_on_uninstall() gate: every network-level FAZ transient deleted on any multisite uninstall regardless of admin choice, violating remove_data_on_uninstall=false contract.

**File:** `uninstall.php:286-312`
**Score:** 85 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per /adamsreview:review reduced-scope flag)

</details>

---

fabiodalez-dev:
> ✅ **11 of 12 findings fixed** (commit `ad5b45d`).
> Live verified on prod:
> - `/faz/v1/banner/en` now emits `Vary: CF-IPCountry` (F111 ✓)
> - 41/41 PR104 specs green post-fix (35.7s)
> 
> **Resolved (11):** F101 F102 F103 F104 F105 F106 F107 F108 F109 F111 F112
> **Skipped (1, audit):** F110 (cosmetic @since — superseded by F109 which removed the parameter entirely)


> ⚠️ **Reduced-scope review** of the last 2 fix commits (5e84413 + 2fb77e5; 13 files, 419 lines).  
> Lanes run: L1 (Sonnet diff-local) + L2 (Opus structural) + L6 (Sonnet security). Phase 4 validation inline-scored.  
> **12 new findings** — most are sub-bugs introduced by the 15 fix commits themselves. Highlights:
> - **F102** (score 90, L2+L6 cross-confirmed): `faz_should_remove_on_uninstall(0)` does NOT check network-level setting on multisite (the F020 fix was wrong)
> - **F103** (85, L2): F016 transaction relies on InnoDB but `install_tables()` doesn't enforce `ENGINE=InnoDB`
> - **F104** (85, L2): F015 microtime cast overflows 32-bit PHP (ClassicPress still supports it)
> - **F105** (85, L2): F008 removed GB from EU in Frontend, but `AMP_Consent::country_in_regions` still includes it
> - **F108** (85, L2): F006/F007's removal of `do_action`+`delete_cache` from create_item leaves no listener notification when update_item fails
> - **F109** (85, L2+L6): F019's `wp_hash($ip, 'nonce')` produces a stable per-salt HMAC — still functionally a PII correlation token
> 
> See per-finding detail below.


### Code review

**Branch:** `feat/multi-banner-geo-routing` → `main`
**PR:** #104 (open)
**Review ID:** `rev_01KRZ7V6EVP19VX1J9QH30AFPM`
**Fix threshold:** not yet set (run `/adamsreview:fix [threshold]` to apply fixes)
**Sub-agent tokens:** 0 across 0 invocations

Found 12 findings across all lanes:
- Deep lane (correctness/security): 8 auto-fixable, 3 partial

---

### ✓ Auto-fixable (11)

| F101 | 80 | correctness | `admin/modules/banners/includes/class-controller.php:327-334` | do_action('faz_after_update_banner') fires on the not-found return path of delete_item (when SELECT … FOR UPDATE returns null), triggering cache-purge hooks (LiteSpeed, WP Rocket, epoch bump) for a no-op DELETE where no banner was actually removed. |
| F102 | 90 | correctness | `uninstall.php:22-33` | faz_should_remove_on_uninstall(0) does NOT check network-level settings. get_blog_option(0,…) is treated by WordPress as get_option() on whichever blog is current at the call site (primary site after the per-blog loop). A multisite admin who set remove_data_on_uninstall=true on subsites but NOT on the primary site will see network-level site_transients survive uninstall. |
| F103 | 85 | correctness | `admin/modules/banners/includes/class-controller.php:85-114` | Transaction-based delete in F016 v2 relies on InnoDB row-level locks, but install_tables()' CREATE TABLE in get_schema() OMITS `ENGINE=InnoDB`. On sites with default_storage_engine=MyISAM (legacy hosts, RDS parameter groups, manual ALTERs), the START TRANSACTION/COMMIT/ROLLBACK/SELECT … FOR UPDATE are silent no-ops. The comment claim 'install_tables() guarantees InnoDB' is incorrect. |
| F104 | 85 | correctness | `admin/modules/banners/includes/class-controller.php:676-685` | F015's microtime epoch overflows 32-bit PHP. (int)(microtime(true)*1000) ≈ 1.75×10^12 today, which exceeds PHP_INT_MAX on 32-bit (2.1×10^9). ClassicPress 1.x still supports 32-bit PHP. The replacement of '$epoch + 1 with PHP_INT_MAX guard' by raw microtime cast silently truncates on 32-bit, collapsing all subsequent invalidations to the same overflowed integer. |
| F105 | 85 | correctness | `frontend/class-amp-consent.php:480-508` | F008 removed GB from EU preset in Frontend::is_country_in_regions, but the AMP mirror in AMP_Consent::country_in_regions still includes 'GB' in 'eu'. Same publisher gets different UK-visitor behaviour on AMP vs regular pages. |
| F106 | 75 | correctness | `includes/class-activator.php:794-799` | F002's `ALTER TABLE ADD COLUMN longtext NOT NULL` without DEFAULT fails on MySQL with sql_mode=STRICT_TRANS_TABLES (default 5.7+) when the table is non-empty. The follow-up UPDATE to backfill never runs because the ALTER errored out, leaving the column missing. |
| F107 | 75 | correctness | `admin/modules/banners/includes/class-controller.php:678-685` | F015 changed autoload=false → autoload=true on faz_banner_cache_epoch. Since this option is written on every banner create/update/delete (hot-write), and is now autoloaded, every save triggers wp_load_alloptions() rebuild on every node. The cross-node propagation goal could be achieved via `updated_option` listener WITHOUT autoload=true. |
| F108 | 85 | correctness | `admin/modules/banners/includes/class-controller.php:194-222` | F006/F007 removed do_action+delete_cache from create_item tail, relying on save()→update_item() to fire them. If update_item early-returns on $updated===false (UPDATE failure), the slug-with-id never persists, the default-flag invariant goes unenforced, and no listener is notified. Pre-fix the explicit tail ran regardless of UPDATE failure. |
| F109 | 85 | security | `includes/class-geolocation.php:227-232` | F019's wp_hash($ip,'nonce') produces a stable per-salt HMAC — third-party filter consumers can use it as a long-term persistent identifier (functionally PII equivalent). The doc's claim 'can still correlate detections of the same client across calls' IS the property that makes hashed IPs GDPR-significant. Fix is debatably worse than raw IP (which is at least clearly PII and treated as such by audit-conscious consumers). |
| F111 | 75 | security | `frontend/modules/banner-rest/class-banner-rest.php:209-221` | F004 made /faz/v1/banner/{lang} CDN-cacheable for 5 min via `Cache-Control: public, max-age=300` WITHOUT a Vary header. If has_country_dependent_banners() returns a stale-false during cache-epoch propagation delay, a CDN can serve a wrong-country banner to visitors in a different region for the 5-minute window. CCPA banner served to EU visitor = compliance issue. |
| F112 | 80 | correctness | `admin/modules/banners/includes/class-controller.php:348-356` | promote_fallback_default() runs AFTER COMMIT — outside the F016 v2 transaction. Two concurrent delete sessions can each read the same fallback candidate and both UPDATE it with banner_default=1, ending with two default rows. The transaction header comment overstates protection: it covers read+delete only, not the post-COMMIT fallback promotion. |

#### F101 — do_action('faz_after_update_banner') fires on the not-found return path of delete_item (when SELECT … FOR UPDATE returns null), triggering cache-purge hooks (LiteSpeed, WP Rocket, epoch bump) for a no-op DELETE where no banner was actually removed.

**File:** `admin/modules/banners/includes/class-controller.php:327-334`
**Score:** 80 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per reduced-scope flag)

#### F102 — faz_should_remove_on_uninstall(0) does NOT check network-level settings. get_blog_option(0,…) is treated by WordPress as get_option() on whichever blog is current at the call site (primary site after the per-blog loop). A multisite admin who set remove_data_on_uninstall=true on subsites but NOT on the primary site will see network-level site_transients survive uninstall.

**File:** `uninstall.php:22-33`
**Score:** 90 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per reduced-scope flag)

#### F103 — Transaction-based delete in F016 v2 relies on InnoDB row-level locks, but install_tables()' CREATE TABLE in get_schema() OMITS `ENGINE=InnoDB`. On sites with default_storage_engine=MyISAM (legacy hosts, RDS parameter groups, manual ALTERs), the START TRANSACTION/COMMIT/ROLLBACK/SELECT … FOR UPDATE are silent no-ops. The comment claim 'install_tables() guarantees InnoDB' is incorrect.

**File:** `admin/modules/banners/includes/class-controller.php:85-114`
**Score:** 85 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per reduced-scope flag)

#### F104 — F015's microtime epoch overflows 32-bit PHP. (int)(microtime(true)*1000) ≈ 1.75×10^12 today, which exceeds PHP_INT_MAX on 32-bit (2.1×10^9). ClassicPress 1.x still supports 32-bit PHP. The replacement of '$epoch + 1 with PHP_INT_MAX guard' by raw microtime cast silently truncates on 32-bit, collapsing all subsequent invalidations to the same overflowed integer.

**File:** `admin/modules/banners/includes/class-controller.php:676-685`
**Score:** 85 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per reduced-scope flag)

#### F105 — F008 removed GB from EU preset in Frontend::is_country_in_regions, but the AMP mirror in AMP_Consent::country_in_regions still includes 'GB' in 'eu'. Same publisher gets different UK-visitor behaviour on AMP vs regular pages.

**File:** `frontend/class-amp-consent.php:480-508`
**Score:** 85 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per reduced-scope flag)

#### F106 — F002's `ALTER TABLE ADD COLUMN longtext NOT NULL` without DEFAULT fails on MySQL with sql_mode=STRICT_TRANS_TABLES (default 5.7+) when the table is non-empty. The follow-up UPDATE to backfill never runs because the ALTER errored out, leaving the column missing.

**File:** `includes/class-activator.php:794-799`
**Score:** 75 (moderate)
**Reason:** inline-scored (Phase 4 validation skipped per reduced-scope flag)

#### F107 — F015 changed autoload=false → autoload=true on faz_banner_cache_epoch. Since this option is written on every banner create/update/delete (hot-write), and is now autoloaded, every save triggers wp_load_alloptions() rebuild on every node. The cross-node propagation goal could be achieved via `updated_option` listener WITHOUT autoload=true.

**File:** `admin/modules/banners/includes/class-controller.php:678-685`
**Score:** 75 (moderate)
**Reason:** inline-scored (Phase 4 validation skipped per reduced-scope flag)

#### F108 — F006/F007 removed do_action+delete_cache from create_item tail, relying on save()→update_item() to fire them. If update_item early-returns on $updated===false (UPDATE failure), the slug-with-id never persists, the default-flag invariant goes unenforced, and no listener is notified. Pre-fix the explicit tail ran regardless of UPDATE failure.

**File:** `admin/modules/banners/includes/class-controller.php:194-222`
**Score:** 85 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per reduced-scope flag)

#### F109 — F019's wp_hash($ip,'nonce') produces a stable per-salt HMAC — third-party filter consumers can use it as a long-term persistent identifier (functionally PII equivalent). The doc's claim 'can still correlate detections of the same client across calls' IS the property that makes hashed IPs GDPR-significant. Fix is debatably worse than raw IP (which is at least clearly PII and treated as such by audit-conscious consumers).

**File:** `includes/class-geolocation.php:227-232`
**Score:** 85 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per reduced-scope flag)

#### F111 — F004 made /faz/v1/banner/{lang} CDN-cacheable for 5 min via `Cache-Control: public, max-age=300` WITHOUT a Vary header. If has_country_dependent_banners() returns a stale-false during cache-epoch propagation delay, a CDN can serve a wrong-country banner to visitors in a different region for the 5-minute window. CCPA banner served to EU visitor = compliance issue.

**File:** `frontend/modules/banner-rest/class-banner-rest.php:209-221`
**Score:** 75 (moderate)
**Reason:** inline-scored (Phase 4 validation skipped per reduced-scope flag)

#### F112 — promote_fallback_default() runs AFTER COMMIT — outside the F016 v2 transaction. Two concurrent delete sessions can each read the same fallback candidate and both UPDATE it with banner_default=1, ending with two default rows. The transaction header comment overstates protection: it covers read+delete only, not the post-COMMIT fallback promotion.

**File:** `admin/modules/banners/includes/class-controller.php:348-356`
**Score:** 80 (strong)
**Reason:** inline-scored (Phase 4 validation skipped per reduced-scope flag)

</details>

---

fabiodalez-dev:
@coderabbitai full review

Recent commits since the last full CodeRabbit review (last was 2026-05-18 09:43 UTC on commit b3e082e):

1. **5e84413** — apply 13 review#1 findings (F001-F020 minus skipped)
2. **2fb77e5** — F010 N+1 refactor + F017 TOCTOU + F016 v2 portable transaction (DELETE … RETURNING was MariaDB-only, switched to START TRANSACTION + SELECT … FOR UPDATE) + phpcs ignore
3. **ad5b45d** — apply 11 review#2 findings (F101-F112): F102 multisite uninstall correct semantics ($faz_any_opted_in accumulator instead of get_blog_option(0)), F103 ENGINE=InnoDB enforced + migration, F104 microtime stored as string (32-bit overflow fix for ClassicPress), F105 AMP EU mirror aligned, F106 3-step nullability migration (NOT NULL strict mode), F107 autoload revert, F108 create_item failure cascade safety net, F109 IP arg removed from consensus filter, F101 do_action gated on $status>0, F111 Vary: CF-IPCountry, F112 promote_fallback inside same transaction
4. **4a863de** — release: bump to 1.14.3
5. **af76b49** — fix(prefcenter): show every visible category even when cookie list is empty (found during remote install test)

Please review with particular attention to:
- Transaction-based delete in **class-controller.php::delete_item()** — promote_fallback_default() now runs INSIDE the transaction
- **$faz_any_opted_in** accumulator pattern in uninstall.php — correct privacy contract?
- **F108 slug-probe heuristic** in create_item — robustness?
- **F104 microtime-as-string** — (string) get_option on legacy int-stored row OK?
- **af76b49 prefcenter fix** — empty-category UX after empty($cookies) check removal

fabiodalez-dev:
> 🔍 **Review #3 (reduced scope)** of commits `2fb77e5..af76b49` (3 commits, 10 files, 319 lines).  
> Lanes run: L1 (Sonnet diff-local) + L2 (Opus structural). L6 skipped — fixes are concurrency/correctness, not security surface.  
> **10 findings** — most concern sub-bugs in the review#2 fixes themselves.  
>
> **CRITICAL (score 90, structural concurrency):**
> - **F301** — promote_fallback_default's internal `delete_cache()` fires PRE-COMMIT inside the F112 transaction → cache poisoning race
> - **F302** — promote_fallback's SELECTs lack `FOR UPDATE` → F112's "extended lock scope" claim is incomplete (REPEATABLE READ ≠ row lock)
> - **F303** — F103 InnoDB enforcement missing on faz_cookies + faz_cookie_categories (also use START TRANSACTION via settings import)
> - **F304** — F102 accumulator misses FAZ_REMOVE_ALL_DATA + empty-network edge case
>
> **MEDIUM/LOW:** F305 (microtime ~100μs resolution, not sub-ms), F306 (sanitize_title idempotency assumption), F307 (F109 BC break docs), F308 (empty-state HTML semantic), F309 (Vary language axis pre-existing), F310 (F101 trade-off observability)


### Code review

**Branch:** `feat/multi-banner-geo-routing` → `main`
**PR:** #104 (open)
**Review ID:** `rev_01KRZWBDZMR7GTQMZDJM5KVZFR`
**Fix threshold:** not yet set (run `/adamsreview:fix [threshold]` to apply fixes)
**Sub-agent tokens:** 0 across 0 invocations

Found 9 findings across all lanes:
- Deep lane (correctness/security): 4 auto-fixable, 3 partial

---

### ✓ Auto-fixable (7)

| F301 | 85 | correctness | `admin/modules/banners/includes/class-controller.php:836` | promote_fallback_default() calls $this->delete_cache() at line 836 INSIDE the F112 transaction (before the outer COMMIT). delete_cache() flushes wp_cache groups and writes faz_banner_cache_epoch. A concurrent request landing between this inner delete_cache and the outer COMMIT misses the cleared cache, DB-fetches the pre-COMMIT snapshot, and re-populates the cache with stale data — defeating the cache invalidation entirely for the full 5-minute TTL window. |
| F302 | 85 | correctness | `admin/modules/banners/includes/class-controller.php:809-824` | promote_fallback_default()'s two SELECTs do NOT use FOR UPDATE. F112's claim that 'the transaction lock scope covers the read+write of the promoted peer too' is incomplete — REPEATABLE READ gives a consistent snapshot, NOT row locks. Two concurrent delete_item transactions can each see the same peer banner_id in their pre-COMMIT snapshots, both call $wpdb->update setting banner_default=1 on the same row (second is a no-op), and concurrently leave zero-defaults if one of the concurrent operations also clears the peer's flag (e.g. another admin updating that peer via update_item's clear_default_on_others). |
| F303 | 85 | correctness | `admin/modules/cookies/includes/class-category-controller.php:96-110` | F103 added ENGINE=InnoDB to faz_banners CREATE TABLE + matching ALTER migration, but faz_cookies and faz_cookie_categories also use START TRANSACTION (in admin/modules/settings/api/class-api.php at lines 718 + 759 for the import flow). Their CREATE TABLE statements DO NOT specify ENGINE=InnoDB. On legacy MyISAM-default hosts, those transactions are silent no-ops — the F103 fix is table-set-incomplete. |
| F304 | 90 | correctness | `uninstall.php:280-307` | F102's $faz_any_opted_in accumulator is set only inside the per-blog foreach. If get_sites() returns empty (degenerate state: pure-network setup, mid-teardown ordering), the loop body never runs, accumulator stays false, network transient sweep is skipped — EVEN IF FAZ_REMOVE_ALL_DATA is defined. Pre-fix `faz_should_remove_on_uninstall(0)` honoured the constant at line 23 first-branch. Gate should be `$faz_force_remove_all || $faz_any_opted_in`. |
| F305 | 75 | correctness | `admin/modules/banners/includes/class-controller.php:737` | F104 comment claims '(string) microtime(true) gives sub-millisecond resolution' but with PHP default precision=14, microtime(true) returns only ~4 decimal digits (~100μs resolution). Two save() calls within ~100μs (common in a single REST that does insert + update + clear_default_on_others) collapse to the same epoch string. Use sprintf('%.6F', microtime(true)) for true microsecond resolution, or (string) hrtime(true) for nanosecond monotonic. |
| F306 | 65 | correctness | `admin/modules/banners/includes/class-controller.php:244-261` | F108 slug-probe compares $persisted_slug !== $slug. sanitize_title is normally idempotent on ASCII alnum + hyphens, but third-party plugins can hook the 'sanitize_title' filter chain to add non-idempotent transforms (emoji-strip + lowercase + custom char strip). In that case the probe fires on EVERY successful create, running the cleanup tail twice. Narrow but real. |
| F309 | 65 | correctness | `frontend/modules/banner-rest/class-banner-rest.php:224-239` | F111 added Vary: CF-IPCountry on the public (non-country-dependent) path but the payload includes the `language` dimension (line 199 — locale-resolved i18n strings). No Vary on language → multilingual sites behind a Vary-aware CDN that respects CF-IPCountry but not the plugin's lang query/cookie share cache entries across languages for the 5-min TTL. Pre-existing exposure, just made more visible by F111. Either include `language` in URL path or add `Vary: Accept-Language`. |

#### F301 — promote_fallback_default() calls $this->delete_cache() at line 836 INSIDE the F112 transaction (before the outer COMMIT). delete_cache() flushes wp_cache groups and writes faz_banner_cache_epoch. A concurrent request landing between this inner delete_cache and the outer COMMIT misses the cleared cache, DB-fetches the pre-COMMIT snapshot, and re-populates the cache with stale data — defeating the cache invalidation entirely for the full 5-minute TTL window.

**File:** `admin/modules/banners/includes/class-controller.php:836`
**Score:** 85 (strong)
**Reason:** inline-scored

#### F302 — promote_fallback_default()'s two SELECTs do NOT use FOR UPDATE. F112's claim that 'the transaction lock scope covers the read+write of the promoted peer too' is incomplete — REPEATABLE READ gives a consistent snapshot, NOT row locks. Two concurrent delete_item transactions can each see the same peer banner_id in their pre-COMMIT snapshots, both call $wpdb->update setting banner_default=1 on the same row (second is a no-op), and concurrently leave zero-defaults if one of the concurrent operations also clears the peer's flag (e.g. another admin updating that peer via update_item's clear_default_on_others).

**File:** `admin/modules/banners/includes/class-controller.php:809-824`
**Score:** 85 (strong)
**Reason:** inline-scored

#### F303 — F103 added ENGINE=InnoDB to faz_banners CREATE TABLE + matching ALTER migration, but faz_cookies and faz_cookie_categories also use START TRANSACTION (in admin/modules/settings/api/class-api.php at lines 718 + 759 for the import flow). Their CREATE TABLE statements DO NOT specify ENGINE=InnoDB. On legacy MyISAM-default hosts, those transactions are silent no-ops — the F103 fix is table-set-incomplete.

**File:** `admin/modules/cookies/includes/class-category-controller.php:96-110`
**Score:** 85 (strong)
**Reason:** inline-scored

#### F304 — F102's $faz_any_opted_in accumulator is set only inside the per-blog foreach. If get_sites() returns empty (degenerate state: pure-network setup, mid-teardown ordering), the loop body never runs, accumulator stays false, network transient sweep is skipped — EVEN IF FAZ_REMOVE_ALL_DATA is defined. Pre-fix `faz_should_remove_on_uninstall(0)` honoured the constant at line 23 first-branch. Gate should be `$faz_force_remove_all || $faz_any_opted_in`.

**File:** `uninstall.php:280-307`
**Score:** 90 (strong)
**Reason:** inline-scored

#### F305 — F104 comment claims '(string) microtime(true) gives sub-millisecond resolution' but with PHP default precision=14, microtime(true) returns only ~4 decimal digits (~100μs resolution). Two save() calls within ~100μs (common in a single REST that does insert + update + clear_default_on_others) collapse to the same epoch string. Use sprintf('%.6F', microtime(true)) for true microsecond resolution, or (string) hrtime(true) for nanosecond monotonic.

**File:** `admin/modules/banners/includes/class-controller.php:737`
**Score:** 75 (moderate)
**Reason:** inline-scored

#### F306 — F108 slug-probe compares $persisted_slug !== $slug. sanitize_title is normally idempotent on ASCII alnum + hyphens, but third-party plugins can hook the 'sanitize_title' filter chain to add non-idempotent transforms (emoji-strip + lowercase + custom char strip). In that case the probe fires on EVERY successful create, running the cleanup tail twice. Narrow but real.

**File:** `admin/modules/banners/includes/class-controller.php:244-261`
**Score:** 65 (moderate)
**Reason:** inline-scored

#### F309 — F111 added Vary: CF-IPCountry on the public (non-country-dependent) path but the payload includes the `language` dimension (line 199 — locale-resolved i18n strings). No Vary on language → multilingual sites behind a Vary-aware CDN that respects CF-IPCountry but not the plugin's lang query/cookie share cache entries across languages for the 5-min TTL. Pre-existing exposure, just made more visible by F111. Either include `language` in URL path or add `Vary: Accept-Language`.

**File:** `frontend/modules/banner-rest/class-banner-rest.php:224-239`
**Score:** 65 (moderate)
**Reason:** inline-scored

</details>

---

fabiodalez-dev:
## 🔎 adamsreview round #4 — sub-bug detection on post-round#3 fixes

**Scope**: `git diff af76b49..0766e52` (the 2 commits that fixed F301-F308 + CodeRabbit#1/#2)  
**Lenses**: L1 (Sonnet, diff-local) + L2 (Opus, structural deep dive)  
**Verdict**: **5 sub-bugs introduced by the round#3 fixes** — 2 HIGH confirmed by both lenses independently, 3 MEDIUM with verified execution paths.

### 🔴 HIGH (consensus L1+L2, confirmed_mechanical)

**R4-S001: F302's `FOR UPDATE` is a no-op when called from `update_item()`** — `admin/modules/banners/includes/class-controller.php:342`  
`promote_fallback_default()`'s docblock says *"MUST run inside the caller's transaction"*. `delete_item()` correctly wraps with `START TRANSACTION/COMMIT`. **`update_item()` does NOT** — it runs in MySQL autocommit mode, so each `SELECT ... FOR UPDATE` acquires the row lock and **releases it immediately at statement end**, before the subsequent `UPDATE`. Two concurrent `update_item()` calls toggling default flags can race exactly like before F302. F302 closed the race for the delete path but **silently left it open for the update path**.  
*Fix shape*: either wrap update_item's update+invariant+promote sequence in a transaction (mirroring delete_item), or move the START TRANSACTION inside promote_fallback_default itself and drop the "caller's transaction" contract.

**R4-S002: CodeRabbit#2 slug-probe success path skips the at-most-one-default invariant tail** — `class-controller.php:267-281`  
When the focused slug UPDATE succeeds (`$patched > 0`), the safety-net tail (`clear_default_on_others` + `delete_cache` + `do_action`) is **entirely skipped**. But `update_item()` failure (the trigger condition) is exactly the scenario where the invariant may have been violated — slug success ≠ invariant satisfied. If the new banner was created with `banner_default=true` and the update_item UPDATE failed, every pre-existing peer still has `banner_default=1` AND no cache flush AND no hook fire.  
*Fix shape*: run the invariant tail unconditionally on `$persisted_slug !== $slug`, then run the focused UPDATE independently. Never gate the invariant on the patch's success.

### 🟡 MEDIUM (single-lens, verified path)

**R4-S003: Sibling-method contract inconsistency on cache invalidation** — `class-controller.php:339-346`  
F301 removed `delete_cache()` from `promote_fallback_default()`. The sibling `clear_default_on_others()` STILL flushes the cache internally. `update_item()` only flushes when `$updated > 0` — but if save() returned 0 rows affected (identical-data UPDATE no-op) and `was_default` was true, `promote_fallback_default` runs (peer's flag changes in DB), but **delete_cache is skipped**. Frontend serves stale "who is the default" answer until next save.

**R4-S004: F303 InnoDB migration has no partial-failure recovery** — `includes/class-activator.php:905-922`  
The 3-table ALTER loop doesn't check `$wpdb->query` return value. ALTER #1 succeeds, ALTER #2 fails (disk pressure, ROW_FORMAT incompatibility on legacy MyISAM rows >8KB), loop continues. `update_db_version` runs unconditionally afterwards, db_version bumps past 1.14.4, the failed table stays MyISAM forever. The exact population this migration targets (legacy mixed-engine hosts) is most likely to hit the failure mode.

**R4-S005: `$wpdb->sitemeta` query on non-multisite when `FAZ_REMOVE_ALL_DATA` is defined** — `uninstall.php:307-321`  
F304 added `$faz_force_remove_all` as a new entry point to the network sitemeta sweep block. On non-multisite installs, `$wpdb->sitemeta` resolves to `wp_sitemeta` which **does not exist** → `$wpdb->get_col` produces a DB error. Pre-existing exposure, but F304 expanded the surface that triggers it.

### Not actionable

L1-R4-S003 (fallback SELECT doesn't filter `banner_default=0`) — idempotent UPDATE on already-default row; just a minor lock-contention smell. L1-R4-S004 (precision drift comment) — documentation accuracy, no runtime bug. L2-R4-S005 (slug-probe race vs concurrent admin edit) — sub-second race window, narrow.

---

**Detection cost**: ~150K tokens (L1 Sonnet + L2 Opus, 2 sub-agents, 2 commits / 283 diff lines). E2E suite running in parallel: 645 tests, currently in progress.

Issue #119, opened 2026-05-20
CI: pr104-review-fixes.spec.ts:288 — GEOIP_COUNTRY_CODE read despite trust filter OFF


fabiodalez-dev:
## Failure (pre-existing on main)

`tests/e2e/specs/pr104-review-fixes.spec.ts:288` fails: same shape as the CF-IPCountry issue but for `HTTP_GEOIP_COUNTRY_CODE` (mod_geoip / nginx-GeoIP shared hosting). With `faz_trust_geoip_country_code` returning `false` (default), `Geolocation::get_visitor_country()` still consumes the header.

Same security concern as the GEO-21 sibling.

## Expected
- Trust filter ON: `GEOIP_COUNTRY_CODE=IT` steers to IT ✅
- Trust filter OFF: result is NOT IT (header ignored) ❌ currently still IT

## Likely root cause
Same code-path as CF-IPCountry — both filters live in `class-geolocation.php`. Fix likely a single conditional move.

## Acceptance
`pr104-review-fixes.spec.ts:288` and related GEOIP_COUNTRY_CODE tests pass.

## Out-of-perimeter from PR #116
Pre-existing on `main`. Cookie-policy-generator branch does not touch geolocation code.

fabiodalez-dev:
Stesso root cause di #118: il mu-plugin di dev `faz-geo-dev-fake-cf.php` registra `__return_true` su `faz_trust_cf_ipcountry_header` e inietta `$_SERVER['HTTP_CF_IPCOUNTRY']='IT'`. Siccome CF-IPCountry viene controllato **prima** di GEOIP nella detection chain di `detect_country()`, la chain risolve `IT` dalla branch CF e non arriva mai alla branch GEOIP che questo test vuole verificare.

Il filtro `faz_trust_geoip_country_code` non è registrato da nessuna parte e ritorna correttamente `false` di default; il gate è quindi corretto in produzione.

Test aggiornato in #122 per:
- stripare entrambi i filtri prima dell'assertion (`remove_all_filters`)
- `unset($_SERVER['HTTP_CF_IPCOUNTRY'], $_SERVER['HTTP_CF_REGION_CODE'])` per evitare lo short-circuit della branch CF

Chiudo come "not a bug, dev env misconfiguration".

Issue #118, opened 2026-05-20
CI: multi-banner-geo-routing.spec.ts:710 (GEO-21) — CF-IPCountry read despite trust filter OFF


fabiodalez-dev:
## Failure (pre-existing on main)

`tests/e2e/specs/multi-banner-geo-routing.spec.ts:710` (test 'GEO-21') fails: with `faz_trust_cf_ipcountry_header` returning `false` (default), `Geolocation::get_visitor_country()` still consumes the spoofed `HTTP_CF_IPCOUNTRY=DE` and steers the result to DE.

This is a **security regression** — the trust filter exists exactly to protect against Host-header spoofing.

## Expected
With trust filter OFF, `CF-IPCountry: DE` MUST NOT alter the resolved country.

## Likely root cause
`includes/class-geolocation.php::get_visitor_country()` reads `$_SERVER['HTTP_CF_IPCOUNTRY']` before checking `apply_filters('faz_trust_cf_ipcountry_header', false)`.

## Acceptance
GEO-21 + GEO-22 pass.

## Out-of-perimeter from PR #116
Pre-existing on `main`. Cookie-policy-generator branch does not touch `class-geolocation.php`.

fabiodalez-dev:
Non è un bug del codice di produzione. `includes/class-geolocation.php` controlla `apply_filters( 'faz_trust_cf_ipcountry_header', false )` come **primo** membro del `&&` in entrambi i call site (linee 93-94 e 155-156). Con il filtro al suo default `false` lo short-circuit evaluation impedisce ogni lettura del header.

Il test E2E falliva perché il dev WP usa il mu-plugin `wp-content/mu-plugins/faz-geo-dev-fake-cf.php` che registra `__return_true` su quel filtro per simulare Cloudflare in locale (senza un vero GeoLite2). Quel mu-plugin **non esiste in produzione** (auto-disabilitato fuori da localhost/WP_DEBUG).

Test aggiornato in #122 per neutralizzare la contaminazione del fixture dev (`remove_all_filters` dentro il `wp eval`). Chiudo come "not a bug, dev env misconfiguration".

Issue #100, opened 2026-05-15
Bug: wp_localize_script payload (`{handle}-js-extra`) can be blocked as analytics — breaks bootstrap (`_fazCfg` / `_fazConfig` undefined before `faz-fw-js-after`)


Myblueroom:
### Summary
With “Alternative asset path” enabled (`faz-fw` handle), the localized store from `wp_localize_script()` is printed as `<script id="faz-fw-js-extra">` containing `var _fazCfg = { … }`.

The script-blocking pipeline (`filter_inline_script_tag` and/or the output-buffer pass in `process_script_tag`) can rewrite **that same** inline tag to `type="text/plain"` and add `data-faz-category="…"` when a provider scanner pattern matches **inside the JSON body** (cookie catalog / vendor strings / URLs are enough for false positives).

Meanwhile `faz-fw-js-before` / `faz-fw-js-after` remain executable `text/javascript`, including the bridge `window._fazConfig = window._fazCfg` and the main bundle that starts with `const _fazStore = window._fazConfig`.

Result: `_fazCfg` never executes → `_fazConfig` is undefined → `TypeError` on `_backupNodes` (or equivalent) and a broken CMP / no `#faz-consent`.

This is a chicken-and-egg: the CMP cannot load because its own localized config script is classified as blocked analytics.

### Environment
- FAZ Cookie Manager: **1.13.16**
- WordPress:  6.9.4
- PHP: 8.1
- Theme: **Bimber** 
- “Alternative asset path”: **enabled**

### Evidence (view-source)
- `id="faz-fw-js-extra"`: `type="text/plain"` + `data-faz-category="analytics"` while containing `var _fazCfg = …`
- `id="faz-fw-js-before"` / `id="faz-fw-js-after"`: `type="text/javascript"` and executed immediately

### Expected
The bootstrap localized payload for FAZ’s own enqueue handles must **never** be neutralised by the scanner/blocker, or it must always execute before the main bootstrap bundle.

### Suggested fix (worked locally)
1. In `filter_inline_script_tag()`: if `$id` ends with `-js-extra` and the derived handle passes `is_own_script_handle()`, return `$tag` unchanged.
2. In `process_script_tag()` (output buffer): same exemption for `<script id="…-js-extra">` owned by FAZ.

Additionally harden the alt-asset bridge to avoid assigning `undefined`:
[...]

PR #122, opened 2026-05-20
test(e2e): unstick 3 pre-existing failures (dev env contamination + design widening)

fabiodalez-dev:
## Summary

Three E2E tests on `main` were red. Investigation showed **none of them was a production bug** — they were assertions that no longer matched the world they ran in.

| Test | Issue | Root cause |
|---|---|---|
| `multi-banner-geo-routing.spec.ts:710` (GEO-21) | #118 | Dev-only mu-plugin `faz-geo-dev-fake-cf` registers `__return_true` on `faz_trust_cf_ipcountry_header` |
| `pr104-review-fixes.spec.ts:288` (GEOIP_COUNTRY_CODE) | #119 | Same mu-plugin contamination (CF chain short-circuits before GEOIP runs) |
| `cache-plugin-autoexclude.spec.ts:138` (#100) | #117 | Code at `frontend/class-frontend.php:3694` was deliberately widened to exempt any plugin's `-js-extra` from substring provider matching; test still asserted the old narrow behavior |

## Why each is not a production bug

**#118 / #119 — Geo trust filters.** `includes/class-geolocation.php` correctly gates both CF-IPCountry and GEOIP_COUNTRY_CODE on their respective `faz_trust_*` filters as the **first** member of the `&&` clause. Short-circuit evaluation means the headers are never read when the filter returns its default `false`. The tests failed because the dev WP install ships a mu-plugin (`wp-content/mu-plugins/faz-geo-dev-fake-cf.php`) that registers `__return_true` on `faz_trust_cf_ipcountry_header` unconditionally on localhost — so the test process inherits a filter that production never has.

Fix: each `wp eval` block now starts with `remove_all_filters('faz_trust_*')` to strip the mu-plugin's contribution. Each `wp eval` is its own PHP process, so this does not leak across tests.

**#117 — `wp_localize_script` exemption.** `is_wp_localize_or_translations_inline_id()` was widened from "own handles only" to "any plugin's `-js-extra` / `-js-translations`" because the substring-match provider detector false-positives on legitimate localized data: trx_addons' `animate_to_mc4wp_form_submitted` matches the `mc4wp` MailChimp pattern, RankMath localizes `addtoany`, Yoast carries `gtag` in instruction strings. Blocking those rewrites the tag to `type="text/plain"`, the `var NAME = …` assignment never runs, and the host page crashes with `ReferenceError: NAME is not defined`. The exemption is plugin-agnostic by core convention (only `wp_localize_script()` produces `-js-extra`; only `wp_set_script_translations()` produces `-js-translations`). The 30-line docstring on the function documents this trade-off in detail.

Fix: the test now asserts the **two-layer contract** instead of the old narrow one:
1. ANY `-js-extra` suffix is exempt (own + foreign both), proving false-positive prevention.
2. A foreign inline script **without** the suffix carrying a provider hit IS still blocked — the cross-plugin consent gate that genuinely matters.

## Test plan

- [x] All 3 previously-red tests pass: `npx playwright test -g 'GEO-21|GEOIP_COUNTRY_CODE is ignored by default|output-buffer fallback never blocks own wp_localize_script payloads'` → `3 passed (9.6s)`
- [x] No regressions in the full spec files: `cache-plugin-autoexclude.spec.ts` + `pr104-review-fixes.spec.ts` → `30 passed (46.3s)`
- [ ] CI green
- [ ] Issues #117, #118, #119 closed via `Closes` lines in the commit body

## Files changed

- `tests/e2e/specs/multi-banner-geo-routing.spec.ts` — GEO-21 strips mu-plugin filter contamination
- `tests/e2e/specs/pr104-review-fixes.spec.ts` — GEOIP test strips CF filter + injected header
- `tests/e2e/specs/cache-plugin-autoexclude.spec.ts` — #100 test asserts the widened two-layer contract

### DIFF
--- a/tests/e2e/specs/cache-plugin-autoexclude.spec.ts
+++ b/tests/e2e/specs/cache-plugin-autoexclude.spec.ts
@@ -136,6 +136,26 @@ test.describe('Cache-plugin auto-exclude (#83 + 1.13.2 post-review)', () => {
   });
 
   test('output-buffer fallback never blocks own wp_localize_script payloads (#100)', async () => {
+    // Two-layer contract enforced by is_wp_localize_or_translations_inline_id():
+    //
+    // 1. ANY inline script whose id ends in `-js-extra` or `-js-translations`
+    //    is exempt from substring provider matching, because WordPress core
+    //    only emits those suffixes for `wp_localize_script()` (data payload)
+    //    and `wp_set_script_translations()` (i18n payload) — neither shape
+    //    hosts executable tracker code. This exemption is plugin-agnostic
+    //    by design: trx_addons localises `animate_to_mc4wp_form_submitted`,
+    //    RankMath includes `addtoany`, Yoast carries `gtag` in instruction
+    //    strings — blocking those breaks the host page with a
+    //    `ReferenceError: NAME is not defined`. See the docstring on
+    //    is_wp_localize_or_translations_inline_id() (frontend/class-frontend.php)
+    //    for the full rationale. A malicious plugin abusing this suffix
+    //    to hide a tracker has the entire rest of the page to exfiltrate
+    //    from — the assumption "core appends `-js-extra` to data only" is
+    //    the same one WordPress itself relies on.
+    //
+    // 2. A foreign inline script WITHOUT the `-js-extra` / `-js-translations`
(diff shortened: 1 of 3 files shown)


fabiodalez-dev:
@coderabbitai full review

Issue #26, opened 2026-03-11
Missing Link color settings


mocana:
I am trying this now on a live site and noticed there are no controls for the link text color. The buttons and borders all seemed to work OK, but I couldn't customize the link text.

<img width="1308" height="980" alt="Image" src="https://github.com/user-attachments/assets/224e1278-f66c-43e5-8970-bc33bac86f4b" />

PR #127, opened 2026-05-27
feat: auto-detect IAB TCF vendors + Cookie Policy third-party services from scanned cookies

fabiodalez-dev:
## Summary

Implements the "auto-scan for ad vendors" feature requested by Niharika in her "Literally Perfect" wp.org review. The 700+ row IAB Global Vendor List is intimidating to anyone who doesn't already know which vendors apply to their stack; this PR closes that gap by reading the cookie scanner's domain inventory, mapping known ad-tech domains to their IAB GVL vendor IDs, and pre-ticking the matching rows on the Vendor List admin page.

End-to-end: cookie scanner → wp_faz_cookies → curated domain map → REST suggestion → admin button → deferred save.

## What lands

**Backend (commit `d15541a`)**

- `admin/modules/gvl/data/domain-to-vendor.json` — curated map of 23 ad-tech domains → 7 IAB GVL vendor IDs (Google, Microsoft, LinkedIn, TikTok, Criteo, Taboola, Outbrain). Every ID was cross-checked against the live GVL (v160) on 2026-05-27; vendors that have left the TCF (Meta, Snap, X/Twitter, Reddit, Hotjar) are intentionally absent so the suggester never proposes a stale ID. Schema versioned, file ships with a `_comment` inviting PRs from the community to extend the dataset.
- `Gvl::suggest_vendor_ids_from_scanned_cookies()` — reads distinct `domain` values from `wp_faz_cookies`, matches each against the map with exact-equality first and then a suffix-with-leading-dot guard (`.notfacebook.com` MUST NOT match `facebook.com`), intersects the result with the currently downloaded GVL so de-registered vendors are filtered out silently.
- `Gvl_Api::suggest_from_cookies()` — REST callback. `GET /faz/v1/gvl/suggest` returns `{ vendor_ids, already_selected, newly_suggested, gvl_available }`. Read-only by design: never mutates `faz_gvl_selected_vendors`. The existing `POST /selected` route remains the single source of truth for persistence.

**Admin UI (commit `c2f5356`)**

- Button "Auto-detect from cookie scan" added next to "Save Selection" on the Vendor List (IAB) admin page. Click triggers the `/suggest` route, merges the suggested IDs into the in-memory selection state, re-renders the visible table so the new ticks appear, and writes an inline status string + toast notification summarising "Pre-ticked N vendor(s), click Save Selection to apply".
- Deferred-save UX: the auto-detection never writes the option on its own. The admin still has to click "Save Selection" to persist. This keeps the action auditable — the admin SEES which boxes got ticked and can untick any false positive before committing.
- Defensive paths: `gvl_available=false` nudges towards "Update GVL Now" (top of page), zero-match toast asks the admin to run the cookie scanner first, REST failure surfaces an error toast and leaves the selection unchanged.

## Tests

| Spec | Tests | Time |
|---|---|---|
| `tests/e2e/specs/gvl-vendor-auto-detect.spec.ts` (new) | 7 | 28.2s |

Coverage:
1. Endpoint shape: 4 documented keys present, correct types
2. Real-vendor surfacing: 3 cookies planted (Google + LinkedIn + TikTok) → 3 expected vendor IDs returned, sorted unique
3. Suffix guard regression: `.notfacebook.com` does NOT trigger any match
4. Subdomain match: `.m.linkedin.com` still matches the `linkedin.com` map entry
5. `already_selected` / `newly_suggested` split semantics
6. Read-only invariant: calling `/suggest` does NOT mutate `faz_gvl_selected_vendors`
7. Zero-match graceful path: empty arrays in response, not an error

The spec uses `wpEval` + `$wpdb->insert` to plant fixture cookies and `afterAll` cleans them up so the test site state is restored.

## Test plan

- [x] PHP lint clean on `class-gvl.php` and `class-api.php`
- [x] `node --check` clean on `gvl.js`
- [x] JSON validation on the data file (23 mappings, 7 unique vendor IDs)
- [x] `gvl-vendor-auto-detect.spec.ts` — 7/7 on Chromium against nginx + PHP-FPM + WP 7.0
- [x] Live REST smoke via `wp eval` — endpoint returns 200 + documented shape
- [x] Manual UI smoke via headless Playwright — button click triggers toast + inline status, deferred save preserved

## Out of scope

- No bump to `Version:` / `Stable tag:` / `FAZ_VERSION` / `readme.txt Stable tag` / `CHANGELOG.md` / `README.md` — this branch is the implementation, the bump belongs in a separate commit on top when we decide to roll it into 1.16.4 (or whichever release ships next after 1.16.3).
- Mapping dataset is intentionally minimal (7 vendors). Subsequent PRs adding entries should cite the IAB GVL registration ID alongside the change so we never regress to the wrong-ID problem that bit the first iteration of this dataset.
- No new admin settings — auto-detect is a one-click action, not a persistent setting. Future iteration could add a "run on every cookie scan completion" toggle if the workflow turns out to merit it.

## Summary by CodeRabbit

* **New Features**
  * Pulsanti "Auto-detect" in Vendor List e Cookie Policy: pre-selezionano vendor/servizi dal cookie scanner, mostrano stato accessibile, badge "Detected", conteggi i18n, lock/race-guard e disabilitazione fino a hydration/completamento.
  * Nuovi endpoint REST in sola lettura e file JSON di mapping domain→vendor/service per le proposte.

* **Accessibility & UX**
  * Annuncio live per salvataggio consenso, stati aria-live e miglior gestione loading dei bottoni.

* **Style**
  * Nuova badge compatta; hover non applicato a pulsanti disabilitati.

* **Tests**
  * Ampia suite E2E e mutex per stabilità DB durante i test.

* **Chores**
  * Aggiornamento versione e changelog.


fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
Thanks. Both findings landed in commit `6a428b8`:

- **Major (`class-gvl.php:343`)** — `discovered = 1` added to the SELECT so manually-added cookies (admin Cookies → Add new, discovered=0 by default) cannot feed false-positive vendor suggestions. The auto-detect feature is about scanner-observed traffic; this aligns the implementation with that intent.
- **Nitpick (`gvl.js:320-379`)** — `autoDetectRequestId` race guard added, captured locally before the network call, both .then and .catch bail out when the captured id no longer matches. Mirrors the previewRequestId pattern used elsewhere in the codebase.

Extended the E2E suite with a regression test for the discovered=0 guard (test #8 in the spec). Full run is 8/8 green in 44.9s.

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
### Code review

**Branch:** `feat/iab-vendor-auto-detect` → `main`
**PR:** #127 (open)
**Review ID:** `rev_01KSN8Y4XNX4RJJ3F2FRF7DJY1`
**Sub-agent tokens:** 1,078,048 across 16 invocations

Found 23 findings across all lanes:
- Deep lane (correctness/security): 2 resolved
- Light lane (ux/policy/architecture): 1 uncertain

---

### ✓ Auto-fixable (2)

| # | Score | Impact | File | Issue | Status |
|---|-------|--------|------|-------|--------|
| F002 | 86 | correctness | `tests/e2e/specs/gvl-vendor-auto-detect.spec.ts:147-156` | Test 3 claims to verify the suffix-match dot-guard with '.notfacebook.com' but 'facebook.com' is not in the domain-to-vendor map, so the assertion passes trivially without exercising the guard. | ✓ fixed and verified (`3e19605`) |
| F004 | 88 | correctness | `admin/assets/js/pages/gvl.js:358-387` | autoDetectFromCookies mutates selectedVendors but never calls updateSelectedCount(); the visible 'Selected: N vendors' counter stays stale until the admin manually toggles another checkbox. | ✓ fixed and verified (`3e19605`) |

#### F002 — Test 3 claims to verify the suffix-match dot-guard with '.notfacebook.com' but 'facebook.com' is not in the domain-to-vendor map, so the assertion passes trivially without exercising the guard.

**File:** `tests/e2e/specs/gvl-vendor-auto-detect.spec.ts:147-156`
**Score:** 86 (strong)

**Evidence:**
- tests/e2e/specs/gvl-vendor-auto-detect.spec.ts:147-156 plants '.notfacebook.com' and asserts r.vendor_ids === [].
- admin/modules/gvl/data/domain-to-vendor.json has 23 keys; facebook.com is intentionally absent (Meta is not a TCF vendor).
- includes/class-gvl.php:383-390 suffix-guard loop only iterates over map keys — with facebook.com absent, the guard branch is never reached for the planted '.notfacebook.com' input. The function returns [] because the foreach never finds a candidate key, NOT because the '.' guard rejected one.
- linkedin.com IS in the map (vendor 804); '.notlinkedin.com' would be a true adversarial input that exercises the dot-guard.

**Approach:** Change the test fixture domain from '.notfacebook.com' to '.notlinkedin.com' so the suffix-guard is actually exercised against a mapped domain (linkedin.com → 804). Update the inline PHP comment in class-gvl.php to mirror.

**Files to modify:**
- `tests/e2e/specs/gvl-vendor-auto-detect.spec.ts` — Replace '.notfacebook.com' with '.notlinkedin.com' in test 3; update test name + inline comment. (why: Currently the assertion passes trivially because facebook.com isn't in the map. Using linkedin.com (mapped to 804) forces the guard to actually run.)
- `includes/class-gvl.php` — Update the comment near line 381 to use the linkedin.com example for consistency. (why: Comment uses misleading 'notfacebook.com / facebook.com' example.)

**Verification:**
- Mutation test: temporarily change class-gvl.php:384 to drop the '.' guard; updated test MUST now fail with vendor_ids containing 804.
- Confirm test 4 (.m.linkedin.com positive case) still passes.

**Edge cases to preserve:**
- ltrim('.', ...) at line 368 strips leading dot; '.notlinkedin.com' becomes 'notlinkedin.com' — suffix check '.linkedin.com' correctly fails.
- Real subdomain '.m.linkedin.com' becomes 'm.linkedin.com' — suffix check '.linkedin.com' correctly succeeds.

**Latest fix attempt (fixrun_01KSNAEXFDBBTYD7PMS920YZEY):** fixed and verified

#### F004 — autoDetectFromCookies mutates selectedVendors but never calls updateSelectedCount(); the visible 'Selected: N vendors' counter stays stale until the admin manually toggles another checkbox.

**File:** `admin/assets/js/pages/gvl.js:358-387`
**Score:** 88 (strong)

**Evidence:**
- admin/assets/js/pages/gvl.js:301 — updateSelectedCount() updates #faz-gvl-selected-count.
- Callers at line 113 (loadSelectedVendors), 179 (checkbox change), 298 (toggleSelectAll) — never called from autoDetectFromCookies.
- Line 363 mutates selectedVendors; line 369 calls loadVendors() which calls renderVendors() (lines 134-227) but renderVendors() does NOT call updateSelectedCount().
- Result: 'Selected: N vendors' text remains stale until the admin manually toggles a checkbox, clicks Select-All, or reloads the page.

**Approach:** Add updateSelectedCount() call inside autoDetectFromCookies, immediately after the forEach merge and before/alongside loadVendors().

**Files to modify:**
- `admin/assets/js/pages/gvl.js` — After the forEach that merges suggested ids into selectedVendors, call updateSelectedCount() to refresh the counter. (why: Mirrors toggleSelectAll's flow at line 298. Without this, the counter lies between auto-detect click and next manual interaction.)

**Verification:**
- E2E: capture #faz-gvl-selected-count text before click + after, assert non-stale.
- Manual: click Auto-detect with a known cookie inventory; counter should immediately reflect the new total.

**Edge cases to preserve:**
- loadVendors() must still run for table re-render — the fix is additive, not replacement.

**Latest fix attempt (fixrun_01KSNAEXFDBBTYD7PMS920YZEY):** fixed and verified

</details>

## Light lane — ux, policy, architecture

| # | Score | Impact | File | Finding | Disposition |
|---|-------|--------|------|---------|-------------|
| F018 | 55 | ux | `admin/assets/js/pages/gvl.js:341-346` | 'No GVL' warning branch does not write to the status span aria-live region; only a transient FAZ.notify toast — screen-reader users who miss the toast receive no persistent indication of why the button did nothing. | uncertain |

## Polish — below threshold, clustered (12)

Below-gate findings (score < 45) that cluster in the same area — not worth surfacing individually, but dense enough that a human pass may catch something the pipeline filtered out.

| # | Score | File | Rough location | Concern |
|---|-------|------|----------------|---------|
| F016 | 12 | `admin/assets/js/pages/gvl.js` | L326-332 | Race-guard comment says stale handlers 'leave UI state to the newer invocation', but if newer invocation fails before resolving, the button stays permanently disabled because the stale-path early return skips FAZ.btnLoading(btn, false). |
| F005 | 30 | `admin/assets/js/pages/gvl.js` | L332-369 | Race guard protects the /suggest promise only; the loadVendors() call fired inside the .then() handler is itself an async REST call NOT guarded by autoDetectRequestId, so a rapid second click can let the first invocation's loadVendors() resolve last and paint stale state. |
| F010 | 22 | `admin/assets/js/pages/gvl.js` | L348-350 | Asymmetric defensive coding: vendor_ids has .filter(n => n > 0) but added/already do not, so a zero ID slipping into newly_suggested would leak through to selectedVendors[0] = true. |
| F020 | 22 | `admin/assets/js/pages/gvl.js` | L352-356 | 'No match' notification tells admin to 'run the cookie scanner first if you have not' but provides no in-page link or button to navigate to the scanner. |
| F019 | 40 | `admin/assets/js/pages/gvl.js` | L363-369 | After auto-detect succeeds, loadVendors() re-renders the table, but if the admin has a search filter active that excludes all newly-ticked vendors, the visible table does not change — no empty-state copy or count update tells the admin vendors were pre-ticked off-screen, making the feature appear broken. |
| F006 | 38 | `admin/assets/js/pages/gvl.js` | L381-385 | Chained .replace('%d',x).replace('%d',y) pattern is fragile under translation: if a translator uses positional placeholders %1$d/%2$d or repeats %d in different order, the substitution silently misformats. |
| F022 | 20 | `admin/assets/js/pages/gvl.js` | L388-392 | Catch handler error message 'Auto-detect failed. Check the cookie scanner and try again.' is generic — conflates network errors, PHP fatal errors, and permission errors without distinguishing them. Admin has no actionable info. |
| F014 | 42 | `includes/class-gvl.php` | L349-352 | SELECT DISTINCT domain query uses phpcs:ignore to suppress PreparedSQL warnings rather than using $wpdb->prepare(). CLAUDE.md line 254 requires unconditional prepare(), though it also accepts phpcs:ignore with motivation for custom-table inevitable cases (line 291). |
| F007 | 28 | `includes/class-gvl.php` | L351 | SELECT … WHERE … AND discovered = 1 assumes the schema migration has already added the discovered column; on installs whose dbDelta has not yet run (fresh activation order edge) the query fatals with 'Unknown column discovered'. Earlier SHOW TABLES guard catches missing TABLE but not missing COLUMN. |
| F008 | 22 | `includes/class-gvl.php` | L365-391 | Suffix-match loop is O(N×M): each unmatched domain iterates every map key. Cookie-table can grow into thousands with active scanning and the map will grow with PRs — this becomes the bottleneck. |
| F011 | 15 | `includes/class-gvl.php` | L399-403 | array_intersect_key with array_flip relies on int-key consistency between $matched and array_flip(array_map('intval',...)); a future regression dropping the (int) cast would silently return empty without any log. No type assertion. |
| F024 | 15 | `includes/class-gvl.php` | L422-461 | load_domain_vendor_map() uses process-lifetime static cache for the JSON mapping file; if the file is replaced or corrupted at runtime, the stale cached mapping is used for the entire PHP-FPM worker lifetime without integrity check. |

## Fix runs

### Run `fixrun_01KSNAEXFDBBTYD7PMS920YZEY` — 2026-05-27T18:42:15Z

- Outcomes: 3 fixed and verified
- Commits: `3e19605`

| Finding | Group | Outcome | phase_9_finding |
|---------|-------|---------|-----------------|
| F002 | FG-1 | ✓ fixed and verified |  |
| F004 | FG-2 | ✓ fixed and verified |  |
| F013 | FG-3 | ✓ fixed and verified |  |

---

fabiodalez-dev:
**Scope extension** (commit `58f2c6c`)

Bridges the same auto-detect pattern into the Cookie Policy generator's Third-party services tab. Root cause for Gooloo's empty Drittanbieter table on https://www.gooloo.de/cookie-richtlinie was UX: cookie scanner + TPS tab were two independent flows with no link.

Added:
- `admin/modules/cookie-policy-generator/data/domain-to-service.json` — curated map (~128 domains → 82 service IDs)
- REST: `GET /faz/v1/cookie-policy/suggest-services` + `GET /faz/v1/cookie-policy/detected-services`
- Admin UI: 'Auto-detect from cookie scan' button + 'Detected' badges next to each checkbox the scanner observed
- E2E spec: 10/10 passing in 47.3s

Pattern mirrors the GVL implementation in this PR: deferred-save (admin reviews pre-ticks and clicks Save), dot-prefix suffix guard, discovered=1 restriction, allowlist enforcement via `sanitize_settings` reuse.

fabiodalez-dev:
@coderabbitai full review

Sweep the latest commits on this branch:
- 58f2c6c: feat(cookie-policy) auto-detect third-party services from cookie scan
- 2807ffb: test(cookie-policy) extend service auto-detect E2E suite to 25 tests
- b74f400: test(cookie-policy) add migration test for pre-existing admin selection (test #26)

Focus areas:
- admin/modules/cookie-policy-generator/api/class-cookie-policy-api.php — new methods suggest_services / detected_services / suggest_services_from_scanned_cookies / cookies_table_has_discovered_rows / load_domain_service_map
- admin/modules/cookie-policy-generator/data/domain-to-service.json — curated map, 128 domain entries
- admin/assets/js/pages/cookie-policy.js — autoDetectServices, Detected badge rendering, parallel init() fetch
- admin/views/cookie-policy.php — Auto-detect button + aria-live status span
- admin/class-admin.php — 7 new i18n keys (svcDetectedBadge, svcAutoDetect*)
- tests/e2e/specs/cookie-policy-service-auto-detect.spec.ts — 26-test suite

Pattern parallels the GVL auto-detect feature already in this PR (commits d15541a + c2f5356 + 6a428b8 + 3e19605). Same dot-prefix suffix guard, discovered=1 restriction, deferred-save UX, allowlist enforcement via sanitize_settings reuse.

fabiodalez-dev:
### Code review

**Branch:** `feat/iab-vendor-auto-detect` → `main`
**PR:** #127 (open)
**Review ID:** `rev_01KSQ30J202ENFJBXXNYVX9YH8`
**Sub-agent tokens:** 1,541,813 across 21 invocations

Found 26 findings across all lanes:
- Deep lane (correctness/security): 2 resolved, 1 uncertain

---

### ✓ Auto-fixable (2)

| # | Score | Impact | File | Issue | Status |
|---|-------|--------|------|-------|--------|
| F005 | 62 | correctness | `includes/class-gvl.php:394` | break inside inner suffix-match loop exits after first matching map entry; if two map keys both suffix-match the same domain, only the first yields vendors. | ✓ fixed and verified (`ecb606d`) |
| F008 | 80 | correctness | `admin/assets/js/pages/gvl.js:36-44` | GVL Auto-detect button bound but NOT disabled during initial hydration, so clicking before loadSelectedVendors() resolves causes that handler to reset selectedVendors={} and wipe just-merged auto-detected vendors — the exact race the cookie-policy.js path guards against. | ✓ fixed and verified (`ecb606d`) |

#### F005 — break inside inner suffix-match loop exits after first matching map entry; if two map keys both suffix-match the same domain, only the first yields vendors.

**File:** `includes/class-gvl.php:394`
**Score:** 62 (moderate)

**Evidence:**
- includes/class-gvl.php suffix loop break;s after the first dot-suffix map key — collects only that key's vendor IDs.
- admin/modules/cookie-policy-generator/api/class-cookie-policy-api.php has the identical suffix-match + break pattern for service IDs.
- domain-to-vendor.json: 23 keys, 0 suffix-overlapping pairs — break is LATENT here.
- domain-to-service.json: 128 keys; 'challenges.cloudflare.com'->[turnstile] and 'cloudflare.com'->[cf] diverge; a cookie on cdn-cgi.challenges.cloudflare.com matches both, break keeps only turnstile, drops cf.
- Both cf and turnstile are in the sanitize_settings allowlist, so the final filter does NOT mask the loss — the break is the sole cause.
- exact-match branch correctly uses continue (one domain = one exact map entry, no overlap possible there).

**Approach:** Drop the `break;` in both suffix-match loops so the loop collects IDs from EVERY dot-suffix-matching map key. $matched is ID-keyed so this de-duplicates for free and is a no-op for the common single-match case. Leave the exact-match continue branches untouched.

**Files to modify:**
- `includes/class-gvl.php` — Remove the break; inside the foreach ($map_keys as $key) suffix loop. (why: Latent correctness: the community-extensible map will drop a second match the moment an overlapping suffix pair is added; keeping both helpers identical avoids divergence.)
- `admin/modules/cookie-policy-generator/api/class-cookie-policy-api.php` — Remove the break; inside the foreach ($map_keys as $key) suffix loop. (why: Active defect: a cookie on a deeper subdomain of challenges.cloudflare.com drops the cf service from suggestions.)

**Verification:**
- Plant a discovered cookie on 'cdn-cgi.challenges.cloudflare.com' and call /suggest-services: before fix returns ['turnstile'], after fix ['cf','turnstile'].
- Confirm single-match domains (.m.linkedin.com, .js.stripe.com) still return exactly one ID.
- Re-run both auto-detect E2E specs to confirm well-formed sorted ID arrays.

**Edge cases to preserve:**
- Dot-prefix guard must remain: 'notlinkedin.com' must NOT match 'linkedin.com'.
- Exact-match branch keeps its continue.
- ID-keyed de-dup stays.
- Allowlist filtering stays after the loop, applied to the larger $matched set.

**Latest fix attempt (fixrun_01KSQ4S9P43733DC5HYHC87FHW):** fixed and verified

#### F008 — GVL Auto-detect button bound but NOT disabled during initial hydration, so clicking before loadSelectedVendors() resolves causes that handler to reset selectedVendors={} and wipe just-merged auto-detected vendors — the exact race the cookie-policy.js path guards against.

**File:** `admin/assets/js/pages/gvl.js:36-44`
**Score:** 80 (strong)

**Evidence:**
- gvl.js FAZ.ready calls loadSelectedVendors() AND binds #faz-gvl-auto-detect click -> autoDetectFromCookies; the button is never set disabled before/during the async load.
- admin/views/gvl.php #faz-gvl-auto-detect has no disabled attribute -> clickable from first paint.
- loadSelectedVendors().then() executes selectedVendors = {} then repopulates only from the saved option -> unconditional reset of shared state.
- autoDetectFromCookies().then() mutates the SAME selectedVendors object; a late loadSelectedVendors resolver wipes any auto-detected id not already saved.
- The existing autoDetectRequestId guard only dedupes two concurrent auto-detect clicks; it has no relationship to the loadSelectedVendors promise.
- cookie-policy.js init() disables #cp-services-auto-detect BEFORE Promise.all and re-enables in the .then() after writeForm — the canonical fix shape gvl.js lacks.

**Approach:** Mirror the cookie-policy.js fix exactly: disable #faz-gvl-auto-detect at FAZ.ready bind time; re-enable it inside loadSelectedVendors() in BOTH the .then() (after repopulate + updateSelectedCount) and the .catch() (so a failed load still leaves it usable). Optionally add a top-of-function `if (!btn || btn.disabled) return;` guard in autoDetectFromCookies().

**Files to modify:**
- `admin/assets/js/pages/gvl.js` — In FAZ.ready, set autoBtn.disabled = true after binding. In loadSelectedVendors() re-enable #faz-gvl-auto-detect in both the .then() (before loadVendors) and the .catch(). Optionally add an early btn.disabled guard at the top of autoDetectFromCookies(). (why: Closes the hydration-vs-autodetect race: the button is inert until the saved selection is loaded, so the late selectedVendors={} reset can never wipe auto-detected ids. Re-enabling in both branches keeps it usable after success or load failure.)
- `admin/views/gvl.php` — (Optional, defense-in-depth) Add disabled to the #faz-gvl-auto-detect button so it is inert before JS hydration; JS re-enables in loadSelectedVendors().then(). Only safe if the JS re-enable lands in the same change. (why: Eliminates the pre-JS click window declaratively, matching the disable-until-ready intent.)

**Verification:**
- Throttle network, click Auto-detect immediately: before fix the success toast shows then the selection resets; after fix the button is disabled until hydration completes and ticks persist.
- Assert button.disabled===true at FAZ.ready and ===false after loadSelectedVendors resolves (both branches).
- Extend the GVL E2E spec to delay GET gvl/selected, click auto-detect early, assert ticks survive + Save persists them.

**Edge cases to preserve:**
- loadSelectedVendors .catch() must re-enable the button (else a fetch error bricks auto-detect).
- Keep the autoDetectRequestId concurrent-click guard.
- Deferred-save semantics: auto-detect only pre-ticks in memory; Save still required.
- If gvl.php gains disabled, the JS re-enable MUST land in the same commit.

**Latest fix attempt (fixrun_01KSQ4S9P43733DC5HYHC87FHW):** fixed and verified

</details>

### ℹ Uncertain (1)

| F011 | 50 | correctness | `admin/assets/js/pages/gvl.js:388-392` | gvl.autoDetectMixed uses two ordered .replace('%d') (vs positional %1$d/%2$d cookie-policy adopted to fix the same fragility); if translated/reordered the two %d swap — the F006 bug. |

Phase 4 couldn't confirm decisively. Re-run `/adamsreview:review` if you suspect this deserves
further investigation with fresh context.

## Polish — below threshold, clustered (11)

Below-gate findings (score < 45) that cluster in the same area — not worth surfacing individually, but dense enough that a human pass may catch something the pipeline filtered out.

| # | Score | File | Rough location | Concern |
|---|-------|------|----------------|---------|
| F003 | 20 | `admin/assets/js/pages/cookie-policy.js` | L415 | scan_available guard uses strict === false so null/undefined/0 fall through to 'No matching' instead of 'Run scanner first'. |
| F001 | 35 | `admin/assets/js/pages/cookie-policy.js` | L443-447 | After positional (%1$d/%2$d) replacements, the plain-%d fallback chain uses newly.length for the first remaining %d and already.length for the second; a translated string mixing positional and plain markers would mis-substitute. |
| F025 | 18 | `admin/assets/js/pages/cookie-policy.js` | L461-464 | Comment says 'Both rendering passes preserve the checkbox state because writeForm() runs after the second render' — the first render creates blank unchecked boxes; state is not preserved during either pass, only restored once by writeForm after both. |
| F012 | 28 | `admin/assets/js/pages/gvl.js` | L355-364 | Backend partitions newly/already against SAVED option but JS merges into in-memory selectedVendors which can diverge (admin toggled without saving); 'N new + M already' counter can misreport vs what admin sees ticked. |
| F032 | 42 | `admin/assets/js/pages/gvl.js` | L356-362 | GVL suggest returns no scan_available field, so empty vendor_ids shows ambiguous 'No matching... Run scanner first if you have not' — conflates 'ran but no match' vs 'never ran' that cookie-policy distinguishes via scan_available===false. |
| F007 | 30 | `admin/assets/js/pages/gvl.js` | L382-384 | autoDetectAllAlready uses suggested.length not already.length; mismatch if any vendor_id filtered by n>0 guard. |
| F022 | 15 | `admin/modules/cookie-policy-generator/api/class-cookie-policy-api.php` | L459-474 | Docblock says response 'mirrors /faz/v1/gvl/suggest for parity' but primary key is 'service_ids' (string[]) here vs 'vendor_ids' (int[]) in GVL — shapes differ in key name and element type. |
| F015 | 20 | `admin/modules/cookie-policy-generator/api/class-cookie-policy-api.php` | L471-512 | suggest_services() and detected_services() each independently call the scan helper AND cookies_table_has_discovered_rows(), so one admin page load re-runs SHOW TABLES + DISTINCT scan + sanitize_settings multiple times; static map cache mitigates file I/O but not repeated DB round-trips. |
| F004 | 15 | `admin/modules/cookie-policy-generator/api/class-cookie-policy-api.php` | L494 | scan_available computed by a second independent DB query after suggest read; if all rows deleted between queries, returns service_ids:[] with scan_available:true. |
| F024 | 20 | `admin/modules/cookie-policy-generator/api/class-cookie-policy-api.php` | L521-522 | Docblock illustrates dot-prefix guard with '.notgoogle.com'/'google.com' but 'google.com' is NOT in domain-to-service.json, referencing a non-existent map entry; inline code comment correctly uses 'notlinkedin.com'/'linkedin.com'. |
| F018 | 30 | `admin/modules/cookie-policy-generator/api/class-cookie-policy-api.php` | L533-534 | phpcs:ignore on SHOW TABLES query lacks '-- motivation' suffix. |

## Fix runs

### Run `fixrun_01KSQ4S9P43733DC5HYHC87FHW` — 2026-05-28T13:10:05Z

- Outcomes: 3 fixed and verified
- Commits: `ecb606d`

| Finding | Group | Outcome | phase_9_finding |
|---------|-------|---------|-----------------|
| F005 | FG-1 | ✓ fixed and verified |  |
| F008 | FG-2 | ✓ fixed and verified |  |
| F010 | FG-3 | ✓ fixed and verified |  |

---

fabiodalez-dev:
### Code review

**Branch:** `feat/iab-vendor-auto-detect` → `main`
**PR:** #127 (open)
**Review ID:** `rev_01KSQC3CBGFJB5CJDX1EBH7C1G`
**Sub-agent tokens:** 804,264 across 10 invocations

Found 10 findings across all lanes:

---

## Fix runs

### Run `fixrun_01KSQEHWS2TBQ08N4Y6AJGYEBD` — 2026-05-28T14:14:13Z

- Outcomes: 1 fixed and verified
- Commits: `eabf0b5`

| Finding | Group | Outcome | phase_9_finding |
|---------|-------|---------|-----------------|
| F007 | FG-1 | ✓ fixed and verified |  |

---

fabiodalez-dev:
### Code review

**Branch:** `feat/iab-vendor-auto-detect` → `main`
**PR:** #127 (open)
**Review ID:** `rev_01KSS9F9TV51HDPTC0C3MW1NJ7`
**Sub-agent tokens:** 587,602 across 18 invocations

Found 20 findings across all lanes:
- Deep lane (correctness/security): 4 resolved

---

### ✓ Auto-fixable (4)

| # | Score | Impact | File | Issue | Status |
|---|-------|--------|------|-------|--------|
| F001 | 58 | correctness | `admin/assets/js/pages/cookie-policy.js:421-448` | When scan_available=true and newly_suggested=[] but already_selected is non-empty, the condition newly.length===0 && already.length===0 is false so execution falls through to display '0 new + N already selected. Click Save to commit.' — the 'Click Save' CTA is misleading because nothing new was auto-detected; gvl.js handles this same case with a distinct 'all already selected' message that omits the save prompt. <sub>(human-confirmed)</sub> | ✓ fixed and verified (`1351700`) |
| F004 | 32 | correctness | `admin/modules/cookie-policy-generator/api/class-cookie-policy-api.php:494-510` | cookies_table_has_discovered_rows() is called as a separate DB operation after suggest_services_from_scanned_cookies() has already executed equivalent domain queries, producing two redundant round-trips per call. A TOCTOU inconsistency is also possible: if scanner rows change between the two calls, scan_available could disagree with service_ids. <sub>(human-confirmed)</sub> | ✓ fixed and verified (`1351700`) |
| F006 | 38 | correctness | `includes/class-gvl.php:398-401` | suggest_vendor_ids_from_scanned_cookies() runs the cookie-domain DB query and full domain→vendor matching even when GVL has not been downloaded; only the final live-ID intersect is skipped, so the /gvl/suggest endpoint returns an UNfiltered vendor_ids list together with gvl_available=false (the JS discards it, but any other consumer of the documented response shape gets stale/unvalidated IDs). <sub>(human-confirmed)</sub> | ✓ fixed and verified (`1351700`) |
| F008 | 20 | correctness | `admin/modules/cookie-policy-generator/api/class-cookie-policy-api.php:589-591` | suggest_services_from_scanned_cookies() launders matched service IDs through sanitize_settings(['third_party_services'=>...]) (returns allowlist order) then sort()s for determinism; but suggest_services() then computes already_selected/newly_suggested with array_intersect/array_diff against $selected WITHOUT re-sorting, so the cookie-policy partition arrays are unsorted while the GVL sibling endpoint sorts both — a parity asymmetry between the two mirrored endpoints. <sub>(human-confirmed)</sub> | ✓ fixed and verified (`1351700`) |

#### F001 — When scan_available=true and newly_suggested=[] but already_selected is non-empty, the condition newly.length===0 && already.length===0 is false so execution falls through to display '0 new + N already selected. Click Save to commit.' — the 'Click Save' CTA is misleading because nothing new was auto-detected; gvl.js handles this same case with a distinct 'all already selected' message that omits the save prompt.

**File:** `admin/assets/js/pages/cookie-policy.js:421-448`
**Score:** 58
**Human-confirmed:** @fabiodalezbackup@gmail.com at 2026-05-29T07:51:09Z — Validator troppo conservativo: la CTA fuorviante è una vera lacuna di parità con gvl.js.
_Promoted from disposition=`uncertain` / actionability=`auto_fixable` / score_phase4=58_
**Fix direction:** Aggiungi in autoDetectServices() (cookie-policy.js) un ramo dedicato per newly.length===0 && already.length>0, speculare al ramo added.length===0 di gvl.js: mostra un messaggio che conferma che tutti i servizi rilevati sono già selezionati e OMETTE il prompt 'Click Save to commit'. Registra una nuova chiave i18n (es. svcAutoDetectAllAlready) accanto alle svcAutoDetect* esistenti. Non toccare gli altri rami.

**Latest fix attempt (fixrun_01KSSBGBSC49YFQ03HSXED611G):** fixed and verified

#### F004 — cookies_table_has_discovered_rows() is called as a separate DB operation after suggest_services_from_scanned_cookies() has already executed equivalent domain queries, producing two redundant round-trips per call. A TOCTOU inconsistency is also possible: if scanner rows change between the two calls, scan_available could disagree with service_ids.

**File:** `admin/modules/cookie-policy-generator/api/class-cookie-policy-api.php:494-510`
**Score:** 32
**Human-confirmed:** @fabiodalezbackup@gmail.com at 2026-05-29T07:51:09Z — Round-trip DB ridondanti verificati su endpoint admin-only; override manuale del verdetto disproven.
_Promoted from disposition=`?` / actionability=`manual` / score_phase4=32_
**Fix direction:** Estrai il check SHOW TABLES LIKE in un helper privato static-cached (es. table_exists()) chiamato sia da suggest_services_from_scanned_cookies() sia da cookies_table_has_discovered_rows(). NON unire le due SELECT: hanno WHERE deliberatamente diversi (domain<>'' vs COUNT non filtrato); unirle romperebbe la semantica di scan_available.

**Latest fix attempt (fixrun_01KSSBGBSC49YFQ03HSXED611G):** fixed and verified

#### F006 — suggest_vendor_ids_from_scanned_cookies() runs the cookie-domain DB query and full domain→vendor matching even when GVL has not been downloaded; only the final live-ID intersect is skipped, so the /gvl/suggest endpoint returns an UNfiltered vendor_ids list together with gvl_available=false (the JS discards it, but any other consumer of the documented response shape gets stale/unvalidated IDs).

**File:** `includes/class-gvl.php:398-401`
**Score:** 38
**Human-confirmed:** @fabiodalezbackup@gmail.com at 2026-05-29T07:51:09Z — Il contratto API non deve esporre vendor_ids non validati; restituire [] quando la GVL non è disponibile.
_Promoted from disposition=`?` / actionability=`auto_fixable` / score_phase4=None_
**Fix direction:** In Gvl::suggest_vendor_ids_from_scanned_cookies(), quando get_data() non ha vendors (GVL non scaricata), restituisci un array vuoto invece del $matched non filtrato, così la risposta documentata non porta mai ID non validati. Mantieni gvl_available=false nella risposta.

**Latest fix attempt (fixrun_01KSSBGBSC49YFQ03HSXED611G):** fixed and verified

#### F008 — suggest_services_from_scanned_cookies() launders matched service IDs through sanitize_settings(['third_party_services'=>...]) (returns allowlist order) then sort()s for determinism; but suggest_services() then computes already_selected/newly_suggested with array_intersect/array_diff against $selected WITHOUT re-sorting, so the cookie-policy partition arrays are unsorted while the GVL sibling endpoint sorts both — a parity asymmetry between the two mirrored endpoints.

**File:** `admin/modules/cookie-policy-generator/api/class-cookie-policy-api.php:589-591`
**Score:** 20
**Human-confirmed:** @fabiodalezbackup@gmail.com at 2026-05-29T07:51:09Z — Asimmetria di parità: gli array partition di cookie-policy non sono ordinati mentre GVL li ordina; aggiungere parità + test.
_Promoted from disposition=`?` / actionability=`auto_fixable` / score_phase4=None_
**Fix direction:** In suggest_services() ordina already_selected e newly_suggested (es. sort()) prima del return, come il gemello GVL. Aggiungi/estendi l'asserzione e2e sull'ordinamento per coprire anche l'endpoint cookie-policy.

**Latest fix attempt (fixrun_01KSSBGBSC49YFQ03HSXED611G):** fixed and verified

</details>

## Polish — below threshold, clustered (8)

Below-gate findings (score < 45) that cluster in the same area — not worth surfacing individually, but dense enough that a human pass may catch something the pipeline filtered out.

| # | Score | File | Rough location | Concern |
|---|-------|------|----------------|---------|
| F005 | 22 | `admin/assets/js/pages/cookie-policy.js` | L413-414 | In the stale-request bail path, the early return fires BEFORE btn.disabled=false, leaving the button permanently disabled if a stale response ever arrives. The btn.disabled guard prevents a second invocation while the first is in-flight, making the stale check unreachable — but if the guard is ever relaxed the stuck-button bug would manifest. |
| F010 | 35 | `admin/assets/js/pages/cookie-policy.js` | L424-437 | autoDetectServices() pre-ticks only checkboxes already present in #cp-services-list; newly_suggested IDs that are NOT among the rendered group catalog (a service in the PHP allowlist+map but missing from renderServicesList groups) are silently dropped with no warning, unlike the GVL path which persists off-page selections; the allowlist and the JS group catalog are kept in sync only by a hand-maintained comment. |
| F019 | 38 | `admin/assets/js/pages/cookie-policy.js` | L449-452 | The 'No matching services' status is shown as kind='ok' (green), giving a success colour to a neutral informational result — colour conflicts with the copy's neutral meaning and may be misread as success. |
| F024 | 28 | `admin/assets/js/pages/cookie-policy.js` | L450-455 | Error message from a network catch block (err.message) is written verbatim into a DOM status element; while textContent is XSS-safe, err.message on a fetch/XHR error can include server-returned response body fragments or internal path information, leaking server-side details to the admin console. |
| F012 | 32 | `admin/assets/js/pages/cookie-policy.js` | L457-465 | Comment says 'Both rendering passes preserve the checkbox state because writeForm() runs after the second render', but if detected.length===0 the second renderServicesList() call is skipped entirely — only one pass occurs, not two. |
| F002 | 28 | `admin/assets/js/pages/gvl.js` | L360-365 | FAZ.btnLoading(btn,true,label) stores btn.textContent into btn.dataset.origText before setting the spinner; if the stale-request guard (requestId !== autoDetectRequestId) fires, the function returns without calling FAZ.btnLoading(btn,false), so origText stays as the spinner text. A subsequent FAZ.btnLoading(btn,true) call from a newer invocation overwrites origText with the spinner text rather than the original button label. |
| F007 | 18 | `admin/assets/js/pages/gvl.js` | L395-405 | autoDetectFromCookies() merges ALL suggested IDs into selectedVendors via data.vendor_ids, but the success message reports added.length + already.length; if the server returns a vendor_ids set that is NOT exactly the disjoint union of newly_suggested ∪ already_selected the counter and the actually-ticked set silently disagree, with no reconciliation against vendor_ids.length. |
| F020 | 22 | `admin/assets/js/pages/gvl.js` | L415-422 | The 'all already selected' message 'All %d auto-detected vendor(s) were already in your selection.' fills %d with suggested.length (total) rather than already.length; correct when suggested===already but would misreport on partial overlap. |

## Fix runs

### Run `fixrun_01KSSBGBSC49YFQ03HSXED611G` — 2026-05-29T08:03:42Z

- Outcomes: 7 fixed and verified
- Commits: `1351700`

| Finding | Group | Outcome | phase_9_finding |
|---------|-------|---------|-----------------|
| F001 | FG-1 | ✓ fixed and verified |  |
| F004 | FG-2 | ✓ fixed and verified |  |
| F006 | FG-3 | ✓ fixed and verified |  |
| F008 | FG-2 | ✓ fixed and verified |  |
| F011 | FG-2 | ✓ fixed and verified |  |
| F015 | FG-1 | ✓ fixed and verified |  |
| F018 | FG-1 | ✓ fixed and verified |  |

---

fabiodalez-dev:
### Code review

**Branch:** `feat/iab-vendor-auto-detect` → `main`
**PR:** #127 (open)
**Review ID:** `rev_01KSSXC1HHS6A1YPWD9E6TQ1NB`
**Sub-agent tokens:** 686,780 across 20 invocations

Found 19 findings across all lanes:
- Deep lane (correctness/security): 1 resolved

---

### ✓ Auto-fixable (1)

| # | Score | Impact | File | Issue | Status |
|---|-------|--------|------|-------|--------|
| F003 | 45 | correctness | `admin/assets/js/pages/cookie-policy.js:508-531` | Promise.all().then() in init() has no outer .catch(); if renderServicesList()/writeForm() throw inside the resolver, autoDetectBtn.disabled=false is never reached and the button is permanently disabled for the session (rejection silently swallowed). <sub>(human-confirmed)</sub> | ✓ fixed and verified (`39fc8da`) |

#### F003 — Promise.all().then() in init() has no outer .catch(); if renderServicesList()/writeForm() throw inside the resolver, autoDetectBtn.disabled=false is never reached and the button is permanently disabled for the session (rejection silently swallowed).

**File:** `admin/assets/js/pages/cookie-policy.js:508-531`
**Score:** 45
**Human-confirmed:** @fabiodalezbackup@gmail.com at 2026-05-29T14:26:24Z — walkthrough: F003 — picked option B (try/finally re-enable)
_Promoted from disposition=`uncertain` / actionability=`auto_fixable` / score_phase4=45_
**Fix direction:** In init() wrap the Promise.all .then() resolver body in try{ ...existing logic... } finally{ if(autoDetectBtn){ autoDetectBtn.disabled = false; } } and remove the standalone autoDetectBtn.disabled=false line. Keep BOTH inner api() .catch() calls intact (settings->null, detected-services->{service_ids:[]}). Preserve the hydration-race guard: writeForm(settings) must still run before the finally re-enable. Mirrors gvl.js re-enable-on-all-paths. Do NOT move the re-enable before Promise.all.

**Latest fix attempt (fixrun_01KST68G0TTXQWFY9PW7AYN8D1):** fixed and verified

</details>

## Polish — below threshold, clustered (8)

Below-gate findings (score < 45) that cluster in the same area — not worth surfacing individually, but dense enough that a human pass may catch something the pipeline filtered out.

| # | Score | File | Rough location | Concern |
|---|-------|------|----------------|---------|
| F011 | 20 | `.github/workflows/plugin-check.yml` | L30-39 | Comment block describes the broken runner as 'ubuntu-24.04' (a specific image) but the job's runs-on is 'ubuntu-latest'; the two diverge whenever GitHub rotates the ubuntu-latest alias, so the comment documents a snapshot that may no longer match the actual job config. |
| F022 | 35 | `.github/workflows/plugin-check.yml` | L53 | CI installs @wordpress/env from npm with no version pin (npm -g i @wordpress/env); on a pull_request trigger this resolves to the latest registry version at run time — a compromised future release executes with the full runner environment. permissions: contents:read limits token scope but not env-var reads or outbound exfiltration. |
| F008 | 28 | `.github/workflows/plugin-check.yml` | L58-65 | The CI gate's correctness rests on the comment-asserted assumption that `wp plugin check ... --categories=plugin_repo` exits non-zero on ERROR findings. The wp-env start step has a `wp cli info` liveness probe (for #579) but the check step has NO assertion that findings were actually evaluated — if a plugin-check version returns 0 with findings, or --categories yields zero checks, the job passes green while shipping Plugin-Check errors. |
| F020 | 25 | `admin/assets/js/pages/cookie-policy.js` | L396-401 | During the scanning/in-progress state the button is disabled and the status announces 'Scanning…' via aria-live, but no aria-busy=true is set on the button — screen-reader users get the text but no machine-readable busy state. Set aria-busy on btnLoading(true) and remove it on btnLoading(false); mirror in gvl.js. |
| F012 | 18 | `admin/assets/js/pages/cookie-policy.js` | L402-403 | Comment says setAutoDetectStatus 'Mirror setStatus(): auto-clear after 3s', but setAutoDetectStatus adds a clearTimeout guard (timer handle) that setStatus does NOT have (setStatus fires a bare setTimeout with no handle). The new function is strictly stronger, not a mirror — the comment understates the divergence. |
| F004 | 30 | `admin/assets/js/pages/cookie-policy.js` | L434-436 | On stale response (myReqId!==autoDetectRequestId) the .then() returns BEFORE re-enabling the button; if the newest in-flight request is abandoned (navigation/network drop) the button stays disabled/spinning with no recovery path. |
| F013 | 15 | `admin/assets/js/pages/cookie-policy.js` | L446-453 | Comment says this branch 'Mirrors gvl.js's autoDetectAllAlready branch'. The structural intent matches but the displayed count differs: gvl.js fills %d with suggested.length (full set) while cookie-policy.js uses already.length (subset). Both internally consistent, but the parity claim overstates the similarity. |
| F007 | 35 | `admin/assets/js/pages/cookie-policy.js` | L508-532 | init() renders the services list twice (sync renderServicesList() then again inside the Promise.all resolver when detected.length>0); the second render rebuilds #cp-services-list DOM, discarding any checkbox the admin manually ticked during the round-trip, and writeForm(settings) re-applies only the SAVED selection — a narrow lost-input window the GVL sibling avoids by not double-rendering from detected-services. |

## Fix runs

### Run `fixrun_01KST68G0TTXQWFY9PW7AYN8D1` — 2026-05-29T15:47:18Z

- Outcomes: 4 fixed and verified
- Commits: `39fc8da`

| Finding | Group | Outcome | phase_9_finding |
|---------|-------|---------|-----------------|
| F003 | FG-1 | ✓ fixed and verified |  |
| F014 | FG-1 | ✓ fixed and verified |  |
| F015 | FG-1 | ✓ fixed and verified |  |
| F017 | FG-2 | ✓ fixed and verified |  |

---

fabiodalez-dev:
### Walkthrough decisions

`rev_01KSSXC1HHS6A1YPWD9E6TQ1NB` · scope=**qualifying** · score_floor=60 · reviewer=fabiodalezbackup@gmail.com · ts=2026-05-29T14:16:32Z

Walking the **Qualifying** scope: of **1** non-auto-eligible finding(s), **1 promoted**, 0 skipped, 0 stopped.

Promoted findings will be picked up by the next `/adamsreview:fix` run via the `human_confirmation` bypass, regardless of its score threshold.

---

#### Promoted

- **F014** — Detected badge uses hard-coded hex (#dff5e1/#1d7d28) diverging from the design tokens · option **A** (Apply auto-recommendation)
  - **Why:** the inline hex don't match the real tokens and `.faz-badge.faz-badge-success` already exists; auto-rec high confidence.
  - **Fix hint:** `className='faz-badge faz-badge-success faz-svc-detected-badge' (keep faz-svc-detected-badge — E2E selector at spec lines 508/511/515), drop inline cssText except margin-left; deliberate token-colour correction; preserve compact sizing via a .faz-svc-detected-badge CSS rule if needed.`
  - **Prior:** disposition=confirmed_manual · score=62

---

Decisions log: this comment is append-only audit — it's never edited in place. Each `/adamsreview:walkthrough` run posts a fresh entry. Current state: see the main review comment and `artifact.md`.

fabiodalez-dev:
### Walkthrough decisions

`rev_01KSSXC1HHS6A1YPWD9E6TQ1NB` · scope=**qualifying** · score_floor=45 · reviewer=fabiodalezbackup@gmail.com · ts=2026-05-29T14:26:24Z

Walking the **Qualifying** scope: of **3** non-auto-eligible finding(s), **3 promoted**, 0 skipped, 0 stopped.

Promoted findings will be picked up by the next `/adamsreview:fix` run via the `human_confirmation` bypass, regardless of its score threshold.

---

#### Promoted

- **F003** — Promise.all().then() in init() has no outer .catch() · option **B** (try/finally re-enable)
  - **Why:** a synchronous throw in renderServicesList()/writeForm() would leave the auto-detect button disabled for the session; gvl.js re-enables on all paths.
  - **Fix hint:** wrap the .then() resolver body in try{…}finally{ autoDetectBtn.disabled=false }, keep both inner api() .catch() calls, preserve writeForm-before-re-enable; do not move re-enable before Promise.all.
  - **Prior:** disposition=uncertain · score=45

- **F015** — 'No matching services' shown green (kind='ok') for a neutral result · option **A** (neutral kind)
  - **Why:** green mis-signals success for a zero-match result; gvl.js treats it as info/neutral.
  - **Fix hint:** change the no-match setAutoDetectStatus(...,'ok') to '' (neutral secondary colour, no auto-clear); don't touch success/error branches; don't add a new kind.
  - **Prior:** disposition=uncertain · score=55

- **F017** — GVL success status never auto-clears (no timer) · option **B** (extract setAutoDetectStatus in gvl.js)
  - **Why:** the success count persists indefinitely; cookie-policy.js auto-clears 'ok' after 3s.
  - **Fix hint:** add a setAutoDetectStatus(msg,kind) helper in gvl.js mirroring cookie-policy.js (timer handle + clearTimeout, 3s, clear only on ok/success; error persists); route success and error paths through it; use faz-gvl-auto-detect-status; do not timer the error path.
  - **Prior:** disposition=uncertain · score=52

---

Decisions log: this comment is append-only audit — it's never edited in place. Each `/adamsreview:walkthrough` run posts a fresh entry. Current state: see the main review comment and `artifact.md`.

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
### Code review

**Branch:** `feat/iab-vendor-auto-detect` → `main`
**PR:** #127 (open)
**Review ID:** `rev_01KSV7EHH8VGWAM07ZFXPSXTV0`
**Sub-agent tokens:** 550,394 across 16 invocations

Found 12 findings across all lanes:
- Deep lane (correctness/security): 2 resolved

---

### ✓ Auto-fixable (2)

| # | Score | Impact | File | Issue | Status |
|---|-------|--------|------|-------|--------|
| F001 | 34 | correctness | `admin/assets/js/pages/gvl.js:37-48` | setAutoDetectStatus's colour ternary only handles 'error'/'ok'/'success'; the noGvl branch passes kind='warning' and noMatch passes kind='info', both falling through to neutral grey (var(--faz-text-secondary)). The 'GVL not downloaded' prerequisite warning is rendered identically to the neutral/scanning state — no visual distinction. <sub>(human-confirmed)</sub> | ✓ fixed and verified (`dc7272f`) |
| F002 | 38 | correctness | `admin/assets/js/pages/cookie-policy.js:514-545` | The init() Promise.all().then(try/finally) has no outer .catch(). The new try/finally re-enables the button but a synchronous throw in renderServicesList()/writeForm() becomes an UNHANDLED promise rejection and the user-visible 'Load failed' status (which main surfaced via the chained .catch on writeForm) is lost. The F003 comment claims finally 'handles' the throw — it only protects the button, not error surfacing. Regression vs main and vs the gvl.js sibling's real .catch. <sub>(human-confirmed)</sub> | ✓ fixed and verified (`dc7272f`) |

#### F001 — setAutoDetectStatus's colour ternary only handles 'error'/'ok'/'success'; the noGvl branch passes kind='warning' and noMatch passes kind='info', both falling through to neutral grey (var(--faz-text-secondary)). The 'GVL not downloaded' prerequisite warning is rendered identically to the neutral/scanning state — no visual distinction.

**File:** `admin/assets/js/pages/gvl.js:37-48`
**Score:** 34
**Human-confirmed:** @fabiodalezbackup@gmail.com at 2026-05-30T01:57:19Z — fix tutto: real warning/info colour gap (4 lenses)
_Promoted from disposition=`?` / actionability=`auto_fixable` / score_phase4=34_
**Fix direction:** In gvl.js setAutoDetectStatus, add a 'warning' branch to the colour ternary: kind==='warning' ? 'var(--faz-warning, #b86900)' between 'error' and the ok/grey branches. Keep 'info' on neutral grey (no --faz-info token exists). Update the comment (~lines 43-44) to state warning/info/scanning are ALL persistent (no timer). Do NOT add a timer to warning/info. Admin JS is NOT in build:min.

**Latest fix attempt (fixrun_01KSV9MMVRGKVCMV98255HG67E):** fixed and verified

#### F002 — The init() Promise.all().then(try/finally) has no outer .catch(). The new try/finally re-enables the button but a synchronous throw in renderServicesList()/writeForm() becomes an UNHANDLED promise rejection and the user-visible 'Load failed' status (which main surfaced via the chained .catch on writeForm) is lost. The F003 comment claims finally 'handles' the throw — it only protects the button, not error surfacing. Regression vs main and vs the gvl.js sibling's real .catch.

**File:** `admin/assets/js/pages/cookie-policy.js:514-545`
**Score:** 38
**Human-confirmed:** @fabiodalezbackup@gmail.com at 2026-05-30T01:57:19Z — fix tutto: restore error surfacing regressed by the F003 try/finally
_Promoted from disposition=`?` / actionability=`auto_fixable` / score_phase4=38_
**Fix direction:** In cookie-policy.js init(), append a trailing .catch() AFTER the Promise.all().then() block that (a) re-enables autoDetectBtn (idempotent with the finally) and (b) surfaces the error via setStatus(t('loadFailed','Load failed') + ': ' + (err && err.message ? err.message : err), 'error'), mirroring gvl.js loadSelectedVendors. Keep the existing try/finally and the two inner api() .catch() calls unchanged.

**Latest fix attempt (fixrun_01KSV9MMVRGKVCMV98255HG67E):** fixed and verified

</details>

## Fix runs

### Run `fixrun_01KSV9MMVRGKVCMV98255HG67E` — 2026-05-30T02:06:42Z

- Outcomes: 5 fixed and verified
- Commits: `dc7272f`

| Finding | Group | Outcome | phase_9_finding |
|---------|-------|---------|-----------------|
| F001 | FG-1 | ✓ fixed and verified |  |
| F002 | FG-2 | ✓ fixed and verified |  |
| F011 | FG-2 | ✓ fixed and verified |  |
| F012 | FG-3 | ✓ fixed and verified |  |
| F013 | FG-2 | ✓ fixed and verified |  |

---

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
### Code review

**Branch:** `feat/iab-vendor-auto-detect` → `main`
**PR:** #127 (open)
**Review ID:** `rev_20260531T092950Z279ca0`
**Sub-agent tokens:** 1,130,790 across 13 invocations

Found 28 findings across all lanes:
- Deep lane (correctness/security): 3 resolved, 1 uncertain

---

### ✓ Auto-fixable (3)

| # | Score | Impact | File | Issue | Status |
|---|-------|--------|------|-------|--------|
| F001 | 55 | correctness | `admin/assets/js/pages/gvl.js:46` | setAutoDetectStatus uses hardcoded hex colours #c4302b (error) and #1d7d28 (success) instead of the --faz-danger / --faz-success design tokens, creating an inconsistency with the sibling cookie-policy.js which was explicitly updated to use the tokens. | ✓ fixed and verified (`ef81258`) |
| F006 | 52 | correctness | `admin/modules/gvl/api/class-api.php:340-355` | suggest_from_cookies() response omits a scan_available key (unlike the cookie-policy sibling), so the JS cannot distinguish 'no GVL' from 'no scan data' from 'no matches'. | ✓ fixed and verified (`5ab9e8a`) |
| F007 | 56 | correctness | `admin/assets/js/pages/gvl.js:416-442` | Auto-detect ticks suggested vendor_ids into the in-memory selectedVendors map but already/newly partition is computed server-side against the SAVED option; a vendor the admin unticked in-session before clicking is silently re-added and reported as 'already selected', overriding the unsaved removal. | ✓ fixed and verified (`5ab9e8a`) |

#### F001 — setAutoDetectStatus uses hardcoded hex colours #c4302b (error) and #1d7d28 (success) instead of the --faz-danger / --faz-success design tokens, creating an inconsistency with the sibling cookie-policy.js which was explicitly updated to use the tokens.

**File:** `admin/assets/js/pages/gvl.js:46`
**Score:** 55

**Latest fix attempt (fixrun_20260531T1300Zfix01):** fixed and verified
- Phase 9 finding: Applied: gvl.js setAutoDetectStatus uses var(--faz-danger)/var(--faz-success) tokens matching cookie-policy.js sibling.

#### F006 — suggest_from_cookies() response omits a scan_available key (unlike the cookie-policy sibling), so the JS cannot distinguish 'no GVL' from 'no scan data' from 'no matches'.

**File:** `admin/modules/gvl/api/class-api.php:340-355`
**Score:** 52

**Latest fix attempt (fixrun_20260531T1330Zfix02):** fixed and verified
- Phase 9 finding: Applied: suggest_vendor_ids_from_scanned_cookies() now returns {vendor_ids, scan_available}; REST + gvl.js gain the third no-scan state. Verified by gvl-vendor-auto-detect spec (scan_available shape + true-on-unmatched).

#### F007 — Auto-detect ticks suggested vendor_ids into the in-memory selectedVendors map but already/newly partition is computed server-side against the SAVED option; a vendor the admin unticked in-session before clicking is silently re-added and reported as 'already selected', overriding the unsaved removal.

**File:** `admin/assets/js/pages/gvl.js:416-442`
**Score:** 56

**Latest fix attempt (fixrun_20260531T1330Zfix02):** fixed and verified
- Phase 9 finding: Applied: gvl.js auto-detect now merges only newly_suggested (added), not the full suggested set — in-session unticks survive. Mirrors the safe cookie-policy.js pattern.

</details>

### Auto-recommendations (2)

_AI-authored fix directions for findings that aren't auto-fixable today. Run `/adamsreview:fix` to batch-apply with one confirmation, or `/adamsreview:walkthrough` to review one-by-one._

| # | Score | Disp | File | Recommendation |
|---|-------|------|------|----------------|
| F018 | 62 | resolved | `admin/assets/js/faz-admin.js:229-249` | _high:_ In FAZ.notify (admin/assets/js/faz-admin.js, the toast-container creation block ~line 229), add role="status" and aria-live="polite" (plu... |
| F019 | 88 | resolved | `admin/assets/css/faz-admin.css:824-835` | _high:_ In admin/assets/css/faz-admin.css, after the existing .faz-toast-success/.faz-toast-error/.faz-toast-info rules (~line 824-835), add: .fa... |

<details><summary>Full recommendations and alternatives</summary>

#### F018 — FAZ.notify toast elements have no ARIA role or aria-live attribute, so screen readers may not announce auto-detect success/error toasts that are dynamically injected.

**File:** `admin/assets/js/faz-admin.js:229-249`
**Score:** 62 (moderate)
**Auto-recommendation (high):** In FAZ.notify (admin/assets/js/faz-admin.js, the toast-container creation block ~line 229), add role="status" and aria-live="polite" (plus aria-atomic="true") to the toast container element when it is created, so dynamically injected toasts are announced to screen readers (WCAG 2.2 SC 4.1.3). A single polite live region on the container is the standard pattern; individual toasts need no per-element role.

**Latest fix attempt (fixrun_20260531T1300Zfix01):** fixed and verified
- Phase 9 finding: Applied: toast container role=status + aria-live=polite; verified by a11y + auto-detect specs in isolation.

#### F019 — The CSS has .faz-toast-success/error/info but no .faz-toast-warning rule, so FAZ.notify(msg, 'warning') calls (GVL not-downloaded and no-match paths) produce an unstyled white-on-white toast.

**File:** `admin/assets/css/faz-admin.css:824-835`
**Score:** 88 (strong)
**Auto-recommendation (high):** In admin/assets/css/faz-admin.css, after the existing .faz-toast-success/.faz-toast-error/.faz-toast-info rules (~line 824-835), add: .faz-toast-warning { background: var(--faz-warning); color: #fff; } so FAZ.notify(msg,'warning') toasts (GVL not-downloaded / no-match paths) render with a visible background instead of unstyled white-on-white. The --faz-warning token already exists.

**Latest fix attempt (fixrun_20260531T1300Zfix01):** fixed and verified
- Phase 9 finding: Applied: .faz-toast-warning CSS rule using --faz-warning token; verified css-custom-properties + auto-detect specs.

</details>

### ℹ Uncertain (1)

| F009 | 58 | correctness | `admin/assets/js/pages/cookie-policy.js:448-472` | newly_suggested/already_selected come from PHP comparing scan matches against the persisted faz_cookie_policy_data option, not the live unsaved checkbox state; a service the admin unticked but has not saved is re-ticked by auto-detect and counted as 'already selected', silently reversing the unsaved change. |

Phase 4 couldn't confirm decisively. Re-run `/adamsreview:review` if you suspect this deserves
further investigation with fresh context.

## Polish — below threshold, clustered (6)

Below-gate findings (score < 45) that cluster in the same area — not worth surfacing individually, but dense enough that a human pass may catch something the pipeline filtered out.

| # | Score | File | Rough location | Concern |
|---|-------|------|----------------|---------|
| F010 | 35 | `admin/assets/js/pages/banner.js` | L186-190 | Contrast-checker hex parser only accepts 6-digit hex (^#?[0-9a-f]{6}$); a valid 3-digit shorthand colour (#fff) returns null and that text/background pair is silently dropped from the WCAG advisory, under-reporting low-contrast pairs. |
| F002 | 35 | `admin/assets/js/pages/banner.js` | L194 | The WCAG sRGB linearisation threshold is 0.03928 (WCAG 2.0 rounded value); the precise WCAG 2.1/2.2 value is 0.04045, producing a slightly incorrect contrast ratio near that boundary. |
| F021 | 25 | `admin/assets/js/pages/banner.js` | L236 | The contrast-warning card uses the hardcoded background color #fff8e6 instead of the design token var(--faz-warning-light, #fff0d5), producing a visually inconsistent tint. |
| F015 | 30 | `includes/class-gvl.php` | L324-329 | The docblock says the function returns an empty array when 'cookies table is empty / scanner never ran / no domain match' but omits a fourth early-return path: when the GVL has not been downloaded (the else branch around line 415). |
| F028 | 30 | `includes/class-gvl.php` | L338-356 | Table name is sanitised via preg_replace whitelist then interpolated into the prepared query; safe against injection, but the pattern strips non-word chars from the prefix so a prefix with a hyphen silently queries a wrong table name without error. |
| F012 | 30 | `includes/class-gvl.php` | L394-416 | GVL suggestion validates IDs only against the live downloaded GVL vendor set, whereas the cookie-policy sibling validates against a fixed allowlist; the two parallel suggestion paths have diverged validation strictness. |

| F003 |  | `frontend/js/script.js:1604-1614` | _fazShowAgeGate's yes callback calls _fazAcceptCookies(pendingChoice) then unconditionally runs _fazRemoveBanner/_fazAfterConsent, whereas the pattern elsewhere checks === false first. |  |

## Fix runs

### Run `fixrun_20260531T1300Zfix01` — 2026-05-31T11:06:56Z

- Outcomes: 3 fixed and verified
- Commits: `ef81258`

| Finding | Group | Outcome | phase_9_finding |
|---------|-------|---------|-----------------|
| F001 | FG-2 | ✓ fixed and verified | Applied: gvl.js setAutoDetectStatus uses var(--faz-danger)/var(--faz-success) tokens matching cookie-policy.js sibling. |
| F018 | FG-1 | ✓ fixed and verified | Applied: toast container role=status + aria-live=polite; verified by a11y + auto-detect specs in isolation. |
| F019 | FG-1 | ✓ fixed and verified | Applied: .faz-toast-warning CSS rule using --faz-warning token; verified css-custom-properties + auto-detect specs. |

### Run `fixrun_20260531T1330Zfix02` — 2026-05-31T11:28:23Z

- Outcomes: 2 fixed and verified
- Commits: `5ab9e8a`

| Finding | Group | Outcome | phase_9_finding |
|---------|-------|---------|-----------------|
| F006 | FG-1 | ✓ fixed and verified | Applied: suggest_vendor_ids_from_scanned_cookies() now returns {vendor_ids, scan_available}; REST + gvl.js gain the third no-scan state. Verified by gvl-vendor-auto-detect spec (scan_available shape + true-on-unmatched). |
| F007 | FG-2 | ✓ fixed and verified | Applied: gvl.js auto-detect now merges only newly_suggested (added), not the full suggested set — in-session unticks survive. Mirrors the safe cookie-policy.js pattern. |

PR #128, opened 2026-05-31
fix(pr127): cookie-policy auto-detect respects in-session manual unticks (F009)

fabiodalez-dev:
## What

Closes the F009 review finding: the Cookie Policy **Auto-detect from cookie scan** re-ticked a detected service the admin had manually **unticked** in-session (before saving), silently reversing their unsaved intent.

The handler already re-ticked only `newly_suggested` (not `already_selected`), but it ignored the admin's live checkbox state.

## Fix (`admin/assets/js/pages/cookie-policy.js`)

- Track service IDs the admin manually unticks this session in a null-prototype set, via a delegated `change` listener on `#cp-services-list`. Programmatic `writeForm()` ticks do not fire `change`, so hydration never pollutes it.
- Auto-detect skips any `newly_suggested` id in that set — a deliberate in-session removal survives, while every untouched service is still auto-ticked.
- The status count reports services **actually** pre-ticked (`newly − skipped`), so it can no longer over-report. New `svcAutoDetectNoneAdded` string for the all-skipped case.
- The set resets on hydration (`writeForm`) and after a successful save — the saved state becomes the new baseline, so a later Auto-detect re-suggests a removed service freely.

## Tests (`cookie-policy-service-auto-detect.spec.ts`)

- **27** — Auto-detect does NOT re-tick a service the admin unticked, but still ticks the ones they left alone.
- **28** — a successful save resets the baseline: a later Auto-detect re-suggests the removed service.

Both use the `cp-auto-*` slug prefix so `beforeEach`'s `resetCookies()` clears them between runs. Full cookie-policy auto-detect spec green (**28/28**); `cookie-policy-integration` green (**10/10**). `POT`/`PO` refreshed.

## Notes

- Branches off `feat/iab-vendor-auto-detect` (depends on the auto-detect feature that lives only there).
- Admin-only JS/i18n change — no frontend impact.

### DIFF
--- a/admin/assets/js/pages/cookie-policy.js
+++ b/admin/assets/js/pages/cookie-policy.js
@@ -36,6 +36,16 @@
 	// one. Mirrors the GVL admin page's autoDetectRequestId (PR #127).
 	var autoDetectRequestId = 0;
 
+	// Service IDs the admin manually UNTICKED during this session (since the
+	// last hydration / save). Auto-detect consults this so a re-run does not
+	// silently re-tick a detected service the admin deliberately removed
+	// before saving (F009). A null-prototype map is used as a Set so a
+	// service id that collides with an Object.prototype member name (e.g.
+	// "constructor") can never read as a spurious truthy hit. Reset on
+	// hydration (writeForm) and after a successful save — the saved state
+	// becomes the new baseline.
+	var userUntickedServices = Object.create(null);
+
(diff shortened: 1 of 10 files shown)


fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai full review

PR #138, opened 2026-06-10
Compliance hardening on 1.17.2: Quebec routing, DNSMPI enforcement, DSAR, a11y, scanner TLS

fabiodalez-dev:
Rebased the compliance work onto current **main (1.17.2)** and addressed the external compliance review. main 1.17.2 already shipped GPC honoring, the 182-day expiry cap, equal-weight buttons, the sale/share split, the server-side CCPA blocker, aggregate-only pageviews and TCF cmp_id gating — those are **not** re-applied here. This PR carries only what main lacks, plus the still-open review findings.

Supersedes #133 (which was branched from the 1.17.1 tag and would have regressed main's newer work).

## Unique fixes carried onto main
- **Quebec Law 25** region routing (resolver Stage 4b + `CA-QC` in `_index.json`) — Quebec visitors were getting federal PIPEDA.
- **Cookie Policy `aria-level`** — `wp_kses_post()` strips `aria-level` while keeping `role="heading"` (axe-critical WCAG 4.1.2). Sanitize with a `post` allowlist extended by `role` + `aria-level`.
- **Trusted-proxy CIDR allowlist** (`faz_ip_in_cidr_list` + `faz_trusted_proxies` gate) — a spoofed `X-Forwarded-For` could rotate past the per-IP throttle.
- US UOOM states (`TX/NJ/MN/MD/NH`) now declare `gpc_honored/gpc_required=true`; **PIPEDA** reclassified to hybrid.
- Removed dead `_template.json` scaffold stubs; test-infra robustness (plugin-lifecycle locator + reseed, setCategoryToggle attach-wait, pr-regression timeouts).

## Review findings addressed
- **[P0] DNSMPI opt-out now enforced** — `get_blocked_categories()` honours the `fazcookie-dnsmpi` cookie: under CCPA/CPRA it blocks the sell/share categories server-side, overriding any prior banner consent (mirrors the Sec-GPC path). Verified via reflection: with the cookie the sold/shared categories block; without it the CCPA notice model leaves them runnable.
- **[P1] DNSMPI raw logs** — `log_optout`/`log_rescind` routed through the central `log_consent()` so the User-Agent is hashed and the URL stripped of query/fragment (was a raw `$wpdb->insert`).
- **[P1] DSAR exporter/eraser/retention** — `faz_privacy_exporter`/`faz_privacy_eraser` now return/delete the requester's `faz_dsar` records by email; daily cron purges DSAR posts older than `faz_settings['dsar']['retention']` months (default 24). Verified end-to-end.
- **[P1] pageview URL minimisation** — store only `scheme://host/path` (drop query/fragment that can hold tokens/PII). (main already removed the stable `faz_sid` identifier.)
- **[P2] scanner TLS** — `sslverify` defaults to verify at all 5 call sites, filterable via `faz_scanner_sslverify` (loopback keeps it off for local self-signed).
- **[P2] composer.json** — reworded "Full … compliance" → "helps you comply … and supports IAB TCF v2.3 (requires a registered CMP ID)".

### Already fixed on main (no action needed)
GPC server-side (Sec-GPC honoured in `get_blocked_categories`), TCF `cmp_id >= 2` gating, `faz_sid` pre-consent identifier removed.

## Tests
- `tests/unit/test-compliance-hardening.php` — **50** standalone assertions, adapted to the 1.17.2 base.
- `tests/e2e/specs/compliance-hardening.spec.ts` — Cookie Policy aria-level (**2** browser checks).
- Validation: PHP lint clean · 48/48 rulesets valid · unit 50/50 · e2e 2/2 · DSAR + DNSMPI verified on the test site.

## Summary by CodeRabbit

* **New Features**
  * Routing regionale con priorità su mapping paese; PIPEDA impostato su modello “hybrid”.
  * Pulizia automatica delle richieste DSAR obsolete.

* **Security**
  * Verifica SSL/TLS abilitata di default (configurabile).
  * Allowlist CIDR per proxy fidati per il riconoscimento header client.

* **Accessibility & Compliance**
  * Migliorata sanitizzazione e ARIA per intestazioni; nuovi test e2e WCAG.

* **Bug Fixes / Behaviour**
  * GPC attivata/obbligatoria in più giurisdizioni; minimizzazione URL salvati; logging consenso esteso (tc_string, gpp_string) e centralizzato per DNSMPI.


fabiodalez-dev:
## Second-review findings — status (commits 12cb9c3, 2bf4de2)

**Fixed (verified):**

| Finding | Fix | Verifica |
|---|---|---|
| [P1] DSAR eraser skips records past batch 1 | always read first page + drain; `done` when batch incomplete | 25 record → **25/25** cancellati (era 20/25) |
| [P1] DNSMPI ignored in GDPR+US "Both" mode + effective only next nav | un-gate the cookie from `law==='ccpa'`; `faz-dnsmpi.js` reloads after opt-out | law=gdpr+cookie → sell/share bloccati; reload immediato |
| [P1] migration overwrites explicit admin classifications | revert to `necessary`-only (categorical); functional/internal seeded at 0 only for **new** installs | — |
| [P2] TLS filter signature inconsistent (PHP 8 ArgumentCountError) | all 5 scanner call sites pass `$url` as 2nd arg | lint OK |
| [P2] DNSMPI log regression `categories=[]` | pass `''` (not `array()`) to `log_consent` | — |
| [P2] PIPEDA grants analytics by default | analytics → `denied-until-action` (CMv2 + default_categories) | ruleset valid, unit 50/50 |
| [P2] audit GPC/DNT empty in real flow | `signal_gpc_received`/`signal_dnt_received` derived server-side from Sec-GPC/DNT headers | `signal_gpc_received=1` verificato |
| [P2] pageview path/title may hold PII | added `faz_pageview_url`/`faz_pageview_title` redaction filters (full path-strip would break analytics) | — |

**Honest residuals (not fixed this round):**

- **[P1] geo rulesets not wired to the runtime banner picker** — accurate, but a **pre-existing architectural** gap on main: the resolved ruleset (Quebec, gpc_required, CMv2) feeds the **consent-log geo audit columns** and the **admin geo API/preview** (so a Quebec visitor's consent log now records `ruleset_id=law25-quebec`), but the live banner is still chosen by `country → target_countries` (admin banner profiles), not by the ruleset. Wiring the ruleset into banner selection + CMv2 application is a large feature beyond this salvage PR; my Quebec change is correct catalog and regresses nothing.
- **[P2] TC/GPP audit strings** still NULL — populating them needs the TCF/GPP frontend modules to forward the encoded strings to the consent endpoint (GPC/DNT are now populated server-side from headers).
- The full-suite **GCM E2E failure** (`ad_storage=denied`) and the **PHPStan baseline (113, mostly WP stub incompatibilities)** are pre-existing on main, unrelated to this branch.

fabiodalez-dev:
## Review (full) — Compliance hardening (Quebec/DNSMPI/DSAR/a11y/scanner TLS + 6 rulesets)

3 lenses (diff-correctness, structural/blast-radius, security). 28 files, all PHP `php -l` clean, all 7 new ruleset JSONs valid + structurally consistent.

**Verdict: mergeable — no HIGH functional blocker.** Strong points confirmed across lenses:
- **DNSMPI enforcement is real, not cosmetic** — `fazcookie-dnsmpi=1` is read server-side in `get_blocked_categories()` and blocks every sell/share category (mirrors the GPC override); opt-out/rescind are recorded centrally via `Controller::log_consent()` (hashed UA, query stripped).
- **Scanner TLS hardened** — all 5 `sslverify:false` replaced with verify-by-default via `apply_filters('faz_scanner_sslverify', …)`, exempting only validated loopback.
- Activator migration safe (guarded, capped, daily cron); privacy exporter/eraser shapes correct (eraser reads page 1 each call → no delete-offset orphaning); resolver signature change backward-compatible.

**Recommended before merge (non-blocking):**
- 🟡 **MED — `tc_string`/`gpp_string` REST params lack `maxLength`** (`consent-logger`): `TEXT` column + unbounded `string` → a leaked nonce could bloat the consent-logs table. Add `'maxLength' => 4096`.
- 🟡 **MED — unit test under-covers the production path**: `test-compliance-hardening.php` calls `resolve()` with 7 positional args (skips `$index_subnational`), exercising the new **Stage 4b** — but production resolves CA-QC via the pre-existing **Stage 3.5** (both call sites pass 9 args). A Stage 3.5/`_regions` regression would pass these tests undetected. Also: Stage 4b + `CA-QC` in `_us_regions` is effectively **dead/redundant** (Stage 3.5 wins) — consider removing or covering it.
- 🟢 **LOW** — `pipeda-canada.json` `_meta.notes` says "implied consent for analytics" but data is `denied-until-action` (data is safer; notes misleading). GPP `__gpp('getGPPData', cb)` assumes sync callback → `gpp_string` audit field may be empty on async CMPs (audit-only). `faz_scanner_sslverify` filter undocumented.

No security HIGH. Clean to merge; the two MED items are quick polish.

PR #144, opened 2026-06-13
fix(geo): align "Geo source not configured" notice with the resolver

fabiodalez-dev:
## Problem

The **Geo Targeting** tab shows an amber *"Geo source not configured"* notice whenever the `maxmind_key` option is empty and the `FAZ_MAXMIND_DB_PATH` constant is unset. But the live resolver (`Geolocation::get_database_path()`) reads the database from `wp-content/uploads/faz-cookie-manager/*.mmdb` — exactly where the plugin's **own GeoLite2 downloader** (and the DB-IP lite fallback) put the file.

So a site that downloaded the database **through the plugin** still saw the warning and assumed geo-targeting was broken, even though detection worked. **False negative.** (This is what a user hit on the support forum: "I do have MaxMind GeoLite2 setup, but it still doesn't allow me to do targeting.")

The inverse was also wrong: `FAZ_MAXMIND_DB_PATH` was read **only** by this notice, never by `get_database_path()` — so defining it silenced the warning while the resolver kept ignoring that file and resolving every visitor to `unknown`. **False positive.**

## Fix

- **`includes/class-geolocation.php`** — `get_database_path()` now honours `FAZ_MAXMIND_DB_PATH` as the first candidate (with precedence over the uploads copies), so the documented constant actually feeds the resolver.
- **`admin/views/banner.php`** — `$faz_has_maxmind` is now derived from `Geolocation::has_database()` (which wraps `get_database_path()`) instead of the inline option/constant check, so the notice mirrors what detection actually uses. A saved `maxmind_key` still counts as configured (the plugin downloads + auto-updates the DB).

Net effect: the notice and the resolver finally agree, and the documented constant works.

## Tests

- New `tests/unit/test-geolocation-db-path.php`: no DB → empty/false; uploaded `GeoLite2-Country.mmdb` (the false-negative case) → resolved/true; `FAZ_MAXMIND_DB_PATH` honoured + precedence. **7/7 pass.**
- Verified live: dropping `GeoLite2-Country.mmdb` into `wp-content/uploads/faz-cookie-manager/` flips `has_database()` to `true` with no fatal in the admin context; removing it flips back. `php -l` clean on both files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Migliorata la rilevazione della disponibilità di MaxMind nel geo-targeting
  * Aggiunta validazione della conformità del database geolocalizzazione per prevenire errori

* **Tests**
  * Aggiunto script automatico per l'esecuzione di test unitari con gestione dei risultati e reporting

### DIFF
--- a/admin/views/banner.php
+++ b/admin/views/banner.php
@@ -751,12 +751,14 @@ defined( 'ABSPATH' ) || exit;
 		// Mirror what the resolver actually uses: Geolocation::has_database()
 		// wraps get_database_path(), which honours FAZ_MAXMIND_DB_PATH and any
 		// .mmdb the plugin downloaded into wp-content/uploads/faz-cookie-manager/.
-		// Previously this notice only checked the maxmind_key option and the
-		// FAZ_MAXMIND_DB_PATH constant inline, so a site whose database had been
-		// downloaded through the plugin still saw "Geo source not configured"
-		// even though geo-detection worked — a false negative. A saved key also
-		// counts as configured (the plugin downloads + auto-updates the DB).
-		$faz_has_maxmind  = ! empty( $faz_geo_settings['geolocation']['maxmind_key'] )
+		// Previously this notice only checked an inline option/constant test, so a
+		// site whose database had been downloaded through the plugin still saw
+		// "Geo source not configured" even though geo-detection worked — a false
+		// negative. A saved license key also counts as configured (the plugin
+		// downloads + auto-updates the DB). The option key is maxmind_license_key
+		// — the value written by Settings → Geolocation (class-settings.php) — not
+		// maxmind_key, which the prior code checked and which is never set.
+		$faz_has_maxmind  = ! empty( $faz_geo_settings['geolocation']['maxmind_license_key'] )
 			|| \FazCookie\Includes\Geolocation::has_database();
 		$faz_has_cf       = (bool) apply_filters( 'faz_trust_cf_ipcountry_header', false );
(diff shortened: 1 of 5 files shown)


fabiodalez-dev:
## Review (full) — Geo "source not configured" notice fix

5 files / 369±, `php -l` clean, adds a unit test. Reviewed the two correctness-critical claims directly.

**Verdict: ready to merge — clean, well-engineered fix.** It corrects three real defects:

1. ✅ **False-negative notice fixed.** The admin "Geo source not configured" check previously read `geolocation.maxmind_key` — a key that is **never written**. Settings actually writes `maxmind_license_key` (verified `class-settings.php:118`). The notice now reads the correct key **and** mirrors what the resolver uses (`Geolocation::has_database()`), so a site that downloaded the DB through the plugin no longer sees a false "not configured" warning.
2. ✅ **`FAZ_MAXMIND_DB_PATH` is now actually used.** It's prepended to the candidate list in `get_database_path()` — previously the constant only silenced the notice but was never read, so an operator pointing it at their own DB still resolved every visitor to "unknown".
3. ✅ **`.mmdb` validation** — `is_valid_mmdb()` reads only the last 128 KB (correct per MMDB spec: metadata marker lives in the tail) and checks for the `\xab\xcd\xefMaxMind.com` marker, with a filesize guard and proper handle close. A corrupt/wrong-format candidate is now skipped so the resolver falls through to the next valid DB instead of throwing on first lookup.

No issues found. Solid.

Issue #146, opened 2026-06-14
Granular consent sub-toggles missing from UI


FINAI-Is-Not-An-Influencer:
## Bug Report: Granular consent sub-toggles missing from UI

**Description**
Hello, thank you for your ongoing work on the plugin. I am opening this issue to report a UI regression that occurred after updating to the latest release. The granular consent options have disappeared from the front-end interface, preventing users from selecting specific services.

**Environment**
* **WordPress Version:** 7.0
* **Plugin Version:** 1.18.2
* **Browser:** Google Chrome Version 148.0.7778.263

**Steps to Reproduce**
1. Update FAZ Cookie Manager to version 1.18.2.
2. Navigate to the front end of the website.
3. Open the cookie preferences modal/banner.
4. Attempt to view the granular services within the main categories.

**Expected Behavior**
The interface should render sub-toggles for specific services under each main category, allowing users to provide or withdraw granular consent. 

**Actual Behavior**
The granular consent sub-toggles no longer appear in the UI and are not clickable. The only interactive elements remaining are the top-level category toggles:
* Necessary
* Functional
* Analytic
* Marketing

**Additional Context**
Thank you for taking the time to review this. Please let me know if you require any browser console logs, DOM inspection details, or further testing to assist with debugging.

fabiodalez-dev:
Resolved in 1.19.0 (merged via #147).

The granular per-service sub-toggles disappeared in 1.18.2 because the per-service / per-cookie opt-in features were intentionally hard-disabled pending a correctness rework (an external review found they did not enforce the granular choice when enabled). That rework is now complete: per-service consent is reintroduced and actually enforced — a denied service is blocked both server-side (pre-consent script block + cookie shredder) and client-side, an explicit service allow overrides a denied category, and the choice persists across reloads and is written to the consent log.

How to get the sub-toggles back:
- Enable **Settings → Per-service consent**. The granular toggles then render under each category in the preference center, sourced from the services actually detected by the cookie scanner on your site (run a scan first if your catalogue is empty). They default to off (opt-in). Turning the option back off cleanly reverts to category-level consent.

Verified end-to-end (preference center renders the per-service sub-toggles under analytics/marketing once services are present). Ships in 1.19.0.

fabiodalez-dev:
This is resolved in 1.19.2, now live on WordPress.org.

[...]

PR #139, opened 2026-06-10
Fix service-level consent: placeholder accept + accordion collapse (#134, #136)

fabiodalez-dev:
Two service-level consent bugs reported against 1.17.2.

## #134 — Placeholder "Accept" accepted the whole category
Clicking "Accept cookies" on a blocked embed (e.g. a YouTube video) accepted the entire **Marketing** category instead of just the **youtube** service.

- `class-placeholder-builder.php`: the button now emits `data-faz-accept-service="<service_id>"` next to `data-faz-accept="<category>"`.
- `script.js`: the delegated handler prefers the service — new `_fazAcceptService(serviceId)` writes `svc.<id>:yes` into the consent cookie and unblocks only that provider, **without** flipping the category to `yes`. Every other provider in the category stays blocked. Falls back to category accept when no service id is present.

## #136 — Service toggle collapsed the accordion
Toggling an individual service inside an expanded category accordion collapsed it. `_fazAttachCategoryListeners` only skipped the **category** switch id; a service toggle had a different id and fell through to `_fazClassToggle()`, which toggled the accordion as a side effect. The listener now short-circuits on any service-toggle / `.faz-switch` / checkbox click and leaves the accordion untouched.

## Tests
`tests/e2e/specs/service-level-consent.spec.ts` (2 passing): the placeholder grants only `svc.youtube` (category not accepted); the shipped runtime carries the accordion guard. `script.min.js` regenerated.

Closes #134
Closes #136

> #135 (granular per-cookie toggles within a service) is a **feature request**, not a bug — tracked separately.

PR #147, opened 2026-06-15
feat(consent): reintroduce per-service consent with the 1.18.2 correctness gaps closed

fabiodalez-dev:
Reintroduces the **per-service consent** control that the 1.18.2 hotfix hard-disabled, this time with the four correctness gaps that motivated the disable resolved. Stacked on `codex/compliance-40-e2e-tests`; the diff is exactly the four per-service commits.

## What changed

**Reintroduction (`1f9480f`)**
- Un-gate the "Enable per-service consent" admin toggle (the per-**cookie** toggle stays gated — its per-cookie revocation cannot be enforced server-side, the service script controls what it writes).
- Re-wire the frontend: `get_store_data()` emits the `_services` list again and `get_service_consent()` honours `svc.*` opt-outs in the cookie-shredder.

**Correctness rework (`617ae70`)**
- **P1-3** — granular `svc.*` decisions are now written to the consent log: the inline logger folds the per-service entries from the consent cookie into the logged `categories` map (GDPR accountability). No schema change.
- **P1-4** — the `fazcookie-consent` cookie is hard-capped at 3500 encoded bytes. When many overrides would exceed it, the lowest-priority entries are dropped deterministically (`ck.*` first, then `svc.*`) with the core category consent always preserved — instead of letting the browser silently drop the whole cookie.
- **P2** — the service toggle list is sourced from the cookies actually **detected** on the site (`wp_faz_cookies`), matched by cookie name with wildcard support (`_ga_*`), instead of the full ~160-provider catalogue.

**Regression coverage (`3756f26`)** — a Playwright spec drives the real shipped code for all three: `_services` is filtered to detected providers, the cookie stays under 4 KB with the core entry preserved, and the `svc.*` decision reaches the consent-log POST.

**Self-review fixes (`55f26b5`)** — a multi-lens review of the above surfaced and this closes:
- **F8** — invalidate the detected-cookie-names transient on cookie writes (was stale up to 6 h after a scan).
- **F5** — sanitize the consent-log `categories` map server-side (cap count/key/value length) so a crafted consent cookie cannot push arbitrary keys/values into the log row.
- **F10** — drop the "nothing scanned yet" fallback that dumped the whole catalogue (over-disclosure + cap pressure); until a scan runs the list is empty and the preference center stays on category-level toggles.
- **F12/F6/F15** — docblock relocation, comment accuracy, and admin help-text guidance ("run a scan first").

## Verification
- `php -l` clean on every touched file.
- Per-service Playwright spec: 3/3 green.
- F8 (transient invalidation) and F5 (categories sanitization) verified end-to-end via WP-CLI on the test site.
- Full pre-existing gate still green: compliance 113/113, unit 11/11.

## Still gated (by design)
Per-**cookie** consent stays behind `if (false)` — its advertised single-cookie revocation is not enforceable server-side (P1-2, architectural). It is kept gated for an easy restore if that enforcement gap is ever closed.

## Summary by CodeRabbit

* **Nuove Funzionalità**
  * Opzione **“Consenso per-servizio”** ora sempre visibile nell’amministrazione e applicata in base ai cookie effettivamente rilevati.
  * Consensi granulari integrati nel consenso e nel ripristino dei contenuti bloccati (inclusi attributi dedicati all’accettazione per servizio).

* **Bug Fixes**
  * Sanitizzazione e validazione più rigorosa dei consensi (limiti e valori consentiti) prima del salvataggio.
  * Enforcement e shredder granulari più affidabili per servizio/cookie, con priorità corrette.

* **Test**
  * Aggiornata ed estesa la suite e2e su consenso per-servizio, enforcement e limiti di dimensione del cookie.


fabiodalez-dev:
@coderabbitai full review

Focus areas for this PR (per-service consent reintroduction + correctness rework):

1. **P1-4 cookie size cap** (`frontend/js/script.js` `_fazSetInStore`): verify a core/category entry is never dropped, that `ck.*` is dropped before `svc.*`, the written value stays < 4096 bytes encoded, and the partition runs only when over budget. Confirm `script.min.js` matches the source.
2. **P2 service list** (`frontend/class-frontend.php` `get_store_data` / `provider_has_detected_cookie` / `get_detected_cookie_names`): only providers with a DETECTED cookie appear; wildcard match (`_ga_*`) anchored; empty list (not the catalogue) when nothing scanned; no ReDoS via the regex pattern side.
3. **Cache coherency** (`__construct` invalidator): `delete_transient('faz_detected_cookie_names')` fires on every cookie-write action and runs in the admin context.
4. **Consent-log sanitization** (`admin/modules/consentlogs/includes/class-controller.php` `log_consent`): a client-writable `fazcookie-consent` cookie cannot push unbounded/arbitrary keys/values into the log row; legitimate `svc.*` decisions still round-trip.
5. **wordpress.org**: output escaping, the `$wpdb->get_col` table-prefix `phpcs:ignore` motivation, no new external calls.

Per-cookie consent intentionally stays gated (`if (false)`) — its single-cookie revocation is not server-side enforceable (architectural).

fabiodalez-dev:
Thanks — good catch on the architectural fragility, addressed in fc8ad9e.

Nuance: the real-world path already worked. The admin UI manages cookies through the `faz/v1` REST API, and the loader (`includes/class-cli.php:181`) explicitly does **not** skip `Frontend` on REST/AJAX requests (`… && ! ( defined('REST_REQUEST') && REST_REQUEST ) && ! self::is_rest_url()`), so the closure was registered when `faz_after_*_cookie` fired inside the REST callback. There is no admin-post (non-REST) cookie-write path today.

That said, tying a cache bust to a conditionally-loaded class is brittle, so I took the suggested fix: both transients (`faz_cookie_scripts_map` and `faz_detected_cookie_names`) are now invalidated by a closure registered **unconditionally** in `CLI::define_public_hooks()`, before the Frontend-skip guard — the plugin's single always-run composition root. The bust now fires on every cookie/category/settings write regardless of request context, and the pre-existing scripts-map invalidation is hardened the same way.

Verified: `has_action('faz_after_create_cookie')` registered and both transients cleared on `faz_after_delete_cookie`; per-service Playwright spec still 3/3.

fabiodalez-dev:
**Review (full, multi-lens) of the per-service enforcement work (#146).**

Core verdict: **the precedence wiring is correct end-to-end.** The structural pass verified PHP↔JS parity for all three cases — service-no + category-yes → both block; service-yes + category-no → both allow; no service choice → both fall to category — plus the shared-pattern "any explicit no wins, else explicit yes wins" on both sides. Reload persistence (the reported bug) is fixed (`_fazRemoveAllDeadCookies` is service-aware and runs after the granular toggles persist), and the detected-cookie-names cache is busted centrally in `Cookie_Controller::delete_cache()`.

🔴 **HIGH — found and fixed (7149ca1):** `_fazApplyGpcOptOut()` ran the per-category cookie shredder **before** clearing the svc.*/ck.* overrides, so a stale `svc.<id>:yes` made the now-service-aware shredder skip cookies a GPC opt-out must delete. GPC is a legally-binding opt-out (CPPA §7025) that overrides any prior consent, including explicit per-service allows — overrides are now cleared **before** the shredder. Verified: GPC suite 2/2, per-service 4/4.

Lower-severity notes (left as intentional / pre-existing, documented here for the record):
- `get_detected_cookie_names()` now filters `discovered = 1`, so manually-added (non-scanned) cookies don't surface a service toggle — matches the "manual = admin curation" precedent, but it's a visible semantics shift.
- The PHP server-side shredder honours `svc.*` but not `ck.*` per-cookie denials (pre-existing; JS covers the live page). Per-cookie consent stays gated anyway.
- Server-blocked tags carry `data-faz-category` but not `data-faz-service`; the service is re-derived client-side via pattern matching on reload (works as long as the pattern is in the block list).
- Deny-wins on a cookie pattern shared by an allowed + a denied service is the documented, intended tradeoff.

Verification on this branch: per-service E2E 4/4 (incl. category-only mode), blocking-compliance 6 passed + 1 env-skip (incl. the reactivated "allow GA, keep Clarity blocked" end-to-end through reload), GPC 2/2, unit logging 16/16, php -l clean, script.min.js in sync.

fabiodalez-dev:
## Review (full) — Per-service consent reintroduction

Most-validated PR of this batch. Multi-lens adversarial review earlier + extensive hands-on verification today (faz-test + live). `php -l` clean.

**Verdict: enforcement is correct and proven end-to-end; one merge blocker (base branch) to resolve.**

**Enforcement — sound (verified):**
- ✅ **Precedence parity PHP↔JS** — explicit service "no" wins → explicit "yes" overrides denied category → else inherit category; identical in the server block, the cookie shredder, and `script.js`.
- ✅ **Reload persistence fixed** — the service-aware shredder runs after the granular toggles persist; `svc.*` survives reload.
- ✅ **GPC opt-out** (HIGH, found + fixed `7149ca1`) — now clears `svc.*` allows **before** the shredder so a legally-binding GPC signal fully overrides explicit service allows.
- ✅ **Cache invalidation** (found + fixed `38ca5b2`) — `get_layout_signature()` now includes `FAZ_VERSION` + per-service/per-cookie flags, so the cached banner template self-heals on plugin update and on toggle (previously a post-update install served stale markup to new `script.js`, silently breaking the sub-toggles — root cause of the #146 symptom). Verified: toggling per-service auto-regenerates with no manual cache clear.
- ✅ **Live end-to-end proof** — with a real third-party service present, the preference center renders the per-service sub-toggles (verified 3 services on the live site via an isolated harness, 18 on the full-data test site), nested under their category accordion, default OFF (opt-in). `faz_store_data` filter added for integration/test extensibility; accordion heading aligned to its toggle.

**Tests:** per-service-consent.spec 4/4 (incl. category-only mode), blocking-compliance 6 pass + 1 env-skip (incl. allow-GA / keep-Clarity end-to-end through reload), unit logging 16/16.

🔴 **Merge blocker (not a code issue): base branch.** This PR targets `codex/compliance-40-e2e-tests` (not `main`, and that branch has no PR). It can't merge to main as-is. Plan (agreed): **rebase the 9 per-service commits directly onto main** so #147 becomes a clean main-based PR. Also note: #139 (placeholder service-accept) depends on this PR being merged first (per the #139 review).

Closes #146. Solid once rebased onto main.

PR #154, opened 2026-06-17
Add missing Czech translation for "Czech" language label

vvvamik:
The cs <option> added to the cookie-policy.php language dropdown had no  Czech translation, so it rendered as "Czech" in the Czech admin UI. Added msgstr "Čeština" and recompiled the .mo.

## Summary by CodeRabbit

* **Nuove Funzionalità**
  * Aggiunto il supporto completo per la lingua ceca nel banner di consenso cookie, con stringhe localizzate per tutte le sezioni GDPR e CCPA, incluse notifiche di consenso, centro preferenze, popup di opt-out e tabella di audit.


fabiodalez-dev:
Thank you, this is a clean catch — the `cs` option was added to the language dropdown without its own Czech translation, so it correctly read "Czech" instead of "Čeština" in the Czech admin UI. The `msgstr "Čeština"` plus the recompiled `.mo` is exactly the right fix.

I'll include it in the next release. Thanks for taking the time to fix the translation and recompile the catalog.

vvvamik:
Added one more file to this PR: cs.json (Czech default banner content for the preference center). Without it, strings like "Always Active" and "Show more/less" fell back to English on Czech installs. Tested locally - the preference center now renders fully in Czech. Happy to split it into a separate PR if you'd prefer.

fabiodalez-dev:
Perfect, please keep it in this same PR — no need to split. It's all Czech localization and the diff stays atomic (cs.json + the .po/.mo catalog), so it's cleaner to review and merge as one unit.

I validated `cs.json` against `en.json`: identical key structure (80/80 keys, nothing missing or extra), valid JSON, and the only value matching English is the audit-table "Cookie" header — which is the same word in Czech, so that's correct. The translations read well (GPC notice and the video placeholder are covered too). The content loader already resolves it automatically since `cs` is in the translated-languages list, so it'll be served on Czech installs with a graceful fallback to English.

Thanks again for the thorough work — testing locally and recompiling the catalog is exactly the right approach. This will ship in the next release.

PR #164, opened 2026-06-26
fix: translate banner chrome (Always Active / audit-table headers) on non-English single-language sites

fabiodalez-dev:
## Problem

Reported for `sk_SK` on wordpress.org (topic *Translation*): after 1.21.0 made "Show more"/"Show less" translatable, three banner strings still rendered in English on the front end despite a translated locale — **"Always Active"**, **"Duration"**, **"Description"** — while "Show more"/"Show less" *did* translate. Reproduced locally in `de_DE` (which ships all four translations).

## Root cause (two distinct bugs)

1. **Audit-table headers were never wrapped.** `Cookie`/`Duration`/`Description` were echoed raw from the banner config (`auditTable.elements.headers`), never routed through gettext, so they could never translate even though they were in the POT (they appear there via the admin UI).

2. **The banner template is built in the wrong locale.** The template is server-rendered and cached (`faz_banner_template`), built by `Template::generate()` which switches the WP locale to `faz_wp_locale( faz_current_language() )`. On a non-multilingual install `faz_current_language()` returns the FAZ **default** language (`'en'`), ignoring the WordPress site locale — so the template, and every gettext string baked into it, was force-built in `en_US`. "Show more"/"Show less" translate because they are resolved at runtime *outside* the cached template, following the WP locale.

## Fix

- `class-shortcodes.php`: new `translate_header()` helper routes the three headers through `translate_default_text()`, plus the matching `switch` cases. Admin-customised header text is preserved (same caveat as the other defaults).
- `class-template.php`: new `resolve_build_locale()` — when the banner language is the un-configured `'en'` on a non-multilingual install and the WordPress site locale is non-English, build in the **WordPress site locale**. The banner *content* still comes from the `'en'` bundle (English copy); only the gettext chrome follows the site language. An explicit non-English FAZ language (e.g. default `de` on a WP `en_US` site) is untouched. The site locale is invariant per site, so the full-page cache stays safe.

## Tests

`tests/e2e/locale-banner-translation.mjs` covers:
- WP `de_DE` + FAZ default `en` → chrome translates (the reported bug) ✓
- WP `en_US` + FAZ default `en` → stays English (no regression) ✓
- WP `en_US` + FAZ explicit `de` → explicit language still wins ✓

Core banner/blocking E2E (`frontend-consent`, `settings-and-blocking`) green (8/8). In the default `en`/`en_US` path `resolve_build_locale()` returns `en_US` — byte-identical to prior behaviour.

## Summary by CodeRabbit

* **Bug Fixes**
  * Migliorata la selezione della lingua usata per generare i banner, soprattutto nei siti WordPress monolingua con lingua di sistema diversa dall’ინგlese.
  * Le intestazioni della tabella di audit cookie ora mostrano in modo più affidabile le traduzioni corrette, anche quando provengono dai valori predefiniti.
  * Risolti casi in cui alcune etichette potevano restare in inglese anziché seguire la lingua attiva o quella di WordPress.

### DIFF
--- a/admin/modules/banners/includes/class-template.php
+++ b/admin/modules/banners/includes/class-template.php
@@ -204,7 +204,7 @@ class Template {
 		// against the runtime locale, not the plugin's language setting.
 		// See GitHub issue tracking banner German-only regression.
 		$locale_switched = false;
-		$target_locale   = function_exists( 'faz_wp_locale' ) ? faz_wp_locale( $this->language ) : '';
+		$target_locale   = $this->resolve_build_locale();
 		if ( $target_locale && function_exists( 'switch_to_locale' ) && $target_locale !== get_locale() ) {
 			$locale_switched = switch_to_locale( $target_locale );
 		}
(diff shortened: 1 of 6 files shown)


fabiodalez-dev:
@coderabbitai full review

PR #166, opened 2026-06-27
feat(#165): Advanced Consent Mode — opt-in, load Google tags before consent

### LINKED ISSUE #165: Advanced Consent Mode: don't hard-block Google tags when GCM is enabled
The GCM module already emits Consent Mode v2 signals, but the script blocker still hard-blocks the Google tag stack until consent — so the plugin only ever does Google's "Basic" mode, never "Advanced". This proposes an opt-in Advanced mode, with the implementation sketched against current `main`.

## What I think it should do

When the GCM module is enabled, I'd expect an **Advanced Consent Mode** option (a simple toggle, default off). When it's on:

- The Google tag stack (gtag.js / GA4 / Google Ads) **loads before consent** with `consent default → denied`, so it sends cookieless/modeled pings, and upgrades to granted on consent. This is Google's "Advanced" mode.
- Everything else — non-Google trackers (Meta, TikTok, LinkedIn, Hotjar…) **and the GTM container** (`gtm.js`) — stays hard-blocked as today.

Default off means existing installs don't change.

## Why

Right now the GCM module emits the Consent Mode signals, but the blocker still converts the gtag stack to `type="text/plain"` until consent. `is_gcm_enabled()` is only used to enqueue `gcm.js` (`frontend/class-frontend.php:348`) — it doesn't affect `get_blocked_categories()` or `process_script_tag()`. So `gcm.js` pushes `consent default → denied` into a `dataLayer` that no live gtag consumes, and **no cookieless pings ever fire**. In practice the plugin only does **Basic** mode (tags after consent), never **Advanced** — even though it already ships the signalling that Advanced needs.

This is the standard model for consent platforms, exposed as a Basic/Advanced switch: Complianz ("Block all Google tags before consent? → No") and CookieYes ("Allow Google tags to fire before consent").

One correctness point from Google's docs: the `consent default` (denied) must be set **synchronously in `<head>` before the gtag snippet** ("default consent status is set too late" is a documented failure mode). `gcm.js` is currently an external `<script src>`, so for Advanced mode the default needs to be emitted inline instead.

The GTM container stays blocked on purpose: it can host non-Google tags that don't read Consent Mode, so exempting it would let them load ungated. Exempt gtag-direct only.

## Proposed implementation

Line numbers are against current `main`.

### 1. Setting — `admin/modules/gcm/includes/class-gcm-settings.php`

`get_defaults()` (line 28), add:

```php
'advanced_mode' => false,
```

`sanitize_option()` (line 212), add `advanced_mode` to the bool group:

```php
case 'status':
case 'advanced_mode':
case 'url_passthrough':
```

New helper:

```php
public function is_advanced_mode() {
    return $this->get( 'status' ) && faz_sanitize_bool( $this->get( 'advanced_mode' ) );
}
```

Plus a checkbox on the GCM page (`admin/views/gcm.php` + `admin/assets/js/pages/gcm.js`) bound to `advanced_mode`.

### 2. Exemption — `frontend/class-frontend.php`, `process_script_tag()`

Just before `set_script_type_plain()` (line 2328):

```php
// Advanced Consent Mode: let Consent-Mode-aware Google tags load; gcm.js
// gates storage instead of hard-blocking them.
if ( $this->gcm_settings->is_advanced_mode()
  && $this->is_gcm_managed_script( $attrs, $content ) ) {
    return $full;
}
```

New helper (gtag-direct only — the `gtm.js` container is intentionally excluded):

```php
private function is_gcm_managed_script( $attrs, $content ) {
    $ctx = $this->get_provider_match_context( $attrs, $content );
    $hay = $ctx['url'] . ' ' . $ctx['content'];
    foreach ( array(
        'googletagmanager.com/gtag/js',
        'gtag(',
        'googleadservices.com/pagead/conversion',
        'googlesyndication.com',
        'doubleclick.net',
    ) as $needle ) {
        if ( false !== stripos( $hay, $needle ) ) {
            return true;
        }
    }
    return false;
}
```

### 3. Client-side mirror — `frontend/js/script.js`

Dynamically injected gtag bypasses the server buffer, so mirror it. Expose the flag near `_gtmDataLayer` (~line 1610):

```php
$store['_gcmAdvanced'] = (bool) $this->gcm_settings->is_advanced_mode();
```

Then early-return `false` in `_fazShouldBlockResource()` / `_fazShouldChangeType()` when `_fazStore._gcmAdvanced && _fazIsGcmManaged(url)`:

```js
function _fazIsGcmManaged(u){
  if(!u) return false; u=String(u).toLowerCase();
  return u.indexOf('googletagmanager.com/gtag/js')>-1
      || u.indexOf('googleadservices.com')>-1
      || u.indexOf('googlesyndication.com')>-1
      || u.indexOf('doubleclick.net')>-1;
}
```

### 4. Synchronous default — `frontend/class-frontend.php`

The `consent default` currently rides on the external `gcm.js` (line 354). In Advanced mode it must be inline, before any Google tag:

```php
add_action( 'wp_head', array( $this, 'print_gcm_default_inline' ), 0 );

public function print_gcm_default_inline() {
    if ( ! $this->gcm_settings->is_advanced_mode() ) { return; }
    // get_gcm_data() (line 1180) already carries the denied signals + wait_for_update.
    $signals = $this->get_gcm_default_signals(); // extracted from the existing payload
    echo "<script>window.dataLayer=window.dataLayer||[];function gtag(){dataLayer.push(arguments);}"
       . "gtag('consent','default'," . wp_json_encode( $signals ) . ");</script>";
}
```

`gcm.js` keeps handling `consent update` on the `fazcookie_consent_update` event — no change there.

## Expected result

With Advanced on: before consent, a `…/collect` request fires with `gcs=G100` and no `_ga` cookie; after consent, `gcs=G111` with `_ga` set. Non-Google trackers and the GTM container stay blocked. With Advanced off, behaviour is identical to today.

Refs: [Google – consent mode](https://developers.google.com/tag-platform/security/guides/consent) · [debugging](https://developers.google.com/tag-platform/security/guides/consent-debugging)

fabiodalez-dev:
Closes #165.

Implements the opt-in **Advanced Consent Mode** for Google Consent Mode v2 proposed in #165 — verified diagnosis, implemented along the reporter's blueprint.

## What it does

A new GCM toggle (default **off**). When on:
- The **gtag-direct** stack (`gtag.js` / GA4 / Google Ads) loads **before consent** with a synchronous `consent default → denied` printed inline in `<head>`, so Google sends cookieless/modeled pings (`gcs=G100`) and upgrades to granted on consent.
- **Everything else stays hard-blocked**: non-Google trackers **and** the **GTM container** (`gtm.js`) — it can host tags that ignore Consent Mode, so exempting it would leak them.
- Default off → existing installs keep doing **Basic** mode (tags only after consent).

## Implementation

| Area | Change |
|------|--------|
| `class-gcm-settings.php` | `advanced_mode` default (false) + bool sanitise + `is_advanced_mode()` |
| `class-frontend.php` | `is_gcm_managed_script()` exemption in `process_script_tag()` (gtag-direct only, gtm.js excluded); `print_gcm_default_inline()` on `wp_head` priority 0 (Google's "set the default synchronously before the gtag snippet" requirement); `advanced_mode` in the `_fazGcm` payload + `_gcmAdvanced` in the store |
| `script.js` | `_fazIsGcmManaged()` client-side mirror in `_fazShouldChangeType` / `_fazShouldBlockResource` for dynamically-injected tags |
| `gcm.js` | skips its own `consent default` in Advanced mode (no duplicate); `consent update` unchanged |
| `gcm.php` | opt-in toggle with an explicit compliance notice |

## Compliance

Loading Google before consent is a site-operator decision, so it's opt-in/default-off with an explicit admin notice (some EU authorities still treat cookieless pings as a third-party transfer).

## Tests

`tests/e2e/gcm-advanced-mode.mjs` — Advanced ON: gtag-direct loads + denied `consent default` inline in `<head>`, `gtm.js` and a non-Google tracker blocked; Advanced OFF: gtag re-blocked, no inline default. **6/6 pass.** `script.min.js` / `gcm.min.js` regenerated.

## Summary by CodeRabbit

* **New Features**
  * Introdotta la modalità **Advanced Consent Mode** per Google Consent Mode v2 con **toggle di opt-in (default: off)** nelle **impostazioni** e nella sezione **Advanced Settings**.
  * In Advanced Mode i **tag Google gestiti** (gtag.js/GA4/Google Ads) possono caricarsi **prima del consenso**, con **`consent default` negato** in `<head>` e aggiornamento dopo l’accettazione; **tracker non-Google** e **gtm.js** restano bloccati (con avviso legale in admin).
  * Aggiornata la **schema API REST** per includere `advanced_mode`.
* **Bug Fixes**
  * Migliorato l’enforcement per evitare blocchi/neutralizzazioni non necessarie sui **tag Google gestiti** in Advanced Mode.
* **Tests**
  * Aggiunto un test E2E di regressione e verificata anche la persistenza dell’opzione via endpoint REST.


fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
Good catch — valid finding, fixed in 938550d. The Advanced Consent Mode exemption now respects an explicit per-service denial: I gated it on `true !== $svc_blocked` in `process_script_tag()`, so a visitor who rejected the specific Google service via per-service consent (`svc.<id>:no`) still gets it hard-blocked; the exemption only applies to the category-default / undecided state. Mirrored client-side — the `_fazShouldChangeType` / `_fazShouldBlockResource` exemptions now run after the explicit `svc.<id>:no` check rather than before it.

fabiodalez-dev:
@coderabbitai full review

Issue #167, opened 2026-06-29
Bricks native lazy-load (data-src / bricks-lazy-hidden) bypasses FAZ blocking for tracker iframes


fabiodalez-dev:
## Summary

On a real **Bricks-theme** site (sandbox `new-website.at`, FAZ build `1.20.0+per-service-reveal`), a Bricks **Video** widget renders a server-side YouTube iframe that **loads before consent and is never blocked** — and, more importantly, the page-wide server-side blocker produces **completely unprocessed HTML** on that page. This is separate from #163 (which is the `<img>`-tile gap for the Leaflet map).

## Evidence (home page, no consent cookie AND with an explicit reject-all cookie)

- The Video element is **server-rendered**: the raw HTTP response contains
  `<div id="brxe-…" class="brxe-video"><iframe loading="lazy" … src="…youtube-nocookie.com/embed/…">` as a **raw iframe** — no `type="text/plain"`, no `data-faz-src`.
- FAZ **knows** the provider: `_providersToBlock` includes `{re: "youtube-nocookie.com/embed", categories: ["marketing"]}` (and `youtube.com/embed`, `ytimg.com`, …). `_block: "1"`, `_showBanner: "1"`, banner renders fine, `fazcookie._diag()` → `ready: true`.
- **`data-faz-src` count in the served HTML = 0.** The output buffer rewrote **nothing on the whole page** — not the video, not anything.
- This holds **even when I set an explicit reject-all `fazcookie-consent` cookie and re-fetch** — so it is not a first-visit / per-visitor effect.

## What I ruled out

- **No full-page cache / optimiser plugin.** Active plugins: Antispam Bee, FAZ Cookie Manager, FluentSMTP, Kadence Security Pro. No LiteSpeed/Rocket/Autoptimize/etc., no page-cache headers.
- **Not a page exclusion.** `script_blocking.excluded_pages` and `banner_control.excluded_pages` are both empty, so `is_blocking_disabled_for_page()` returns false.
- **Not a provider-catalogue miss.** youtube-nocookie + openstreetmap are both present and categorised.
- **Not AMP / admin / ajax.**

## Where it points

`Frontend::process_output_buffer()` returns the HTML unprocessed. With caching, exclusions and provider-knowledge ruled out, the remaining candidates are:

[...]

Issue #163, opened 2026-06-26
Bricks Map / Leaflet (OpenStreetMap) tiles not blocked — img-tile enforcement gap for map providers


fabiodalez-dev:
## Summary

Bricks Builder **Map** element using **Leaflet / OpenStreetMap** renders its map tiles even when consent is denied. Reported by a user (3DRZ, via email) alongside a re-report of the Bricks Video element — the Video path is already handled (#87 / #134), but the Map is a genuine enforcement gap.

## Root cause

FAZ blocks resources at three layers, none of which intercept `<img>`:

1. `document.createElement` override — **only `<script>`** (`frontend/js/script.js:2888` returns early for any non-script node).
2. `_fazMutationObserver` — gates `<script[src]>` / `<iframe[src]>` (incl. nested in page-builder wrappers).
3. Network wrappers (`frontend/js/script.js:3850`) — `fetch`, `XMLHttpRequest`, `sendBeacon`, `WebSocket`.

A Leaflet map draws itself by loading tiles as **plain `<img>` requests** injected at runtime by its JS, e.g. `https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png`. These never pass through `the_content` (so they're never rewritten to `img[data-faz-src]`), and `<img>` creation/src is not gated by any of the three layers. Result: tiles load regardless of consent.

`openstreetmap` **is** already a known provider (`includes/data/known-providers.json`, category `functional`, patterns `openstreetmap.org` / `osm/js/osm`) — the catalogue entry exists, but there is **no enforcement path for image tiles**.

## Distinct from prior Bricks work

- #87 / #91 (shipped 1.13.8) — Bricks Video iframe inside `.brxe-video`: MutationObserver descent + 4-step placeholder. ✅
- #134 / #146 — Bricks lightbox `<a data-pswp-video-url>` click-time interceptor. ✅
- **This issue** — `<img>`-tile providers (Leaflet/OSM, and by extension any tile/image-based map or asset). New scope.

## Proposed approaches (to evaluate)

1. **Gate `<img>` for matching blocked providers.** Extend the runtime interception to `<img>` whose `src` matches a known provider in a *denied* category — block the load, show a placeholder, restore on consent. Must be scoped tightly (only matching provider URLs) to avoid per-image perf cost and false positives.
[...]

PR #168, opened 2026-06-30
fix: block runtime-injected resources before consent — map tiles (#163), Bricks lazy embeds (#167), Web Font Loader stylesheets

fabiodalez-dev:
## Summary

Closes the class of bug where a third party sets a **resource URL at runtime**, after the HTML the server-side blocker saw, so the resource loads (and the visitor's IP reaches the third party) with consent denied. Three reported vectors, one mechanism:

- **#163 — Map tiles as ``.** Leaflet/OpenStreetMap and the Bricks Map element draw the map by assigning tile URLs to `` at runtime.
- **#167 — Bricks lazy-loaded embeds.** Bricks' native lazy-load parks an embed URL in `data-src` and later does `iframe.src = data-src` when the element scrolls into view.
- **Web Font Loader stylesheets** (reported on wp.org for Google Fonts). `webfont.js` creates a `<link>` and sets its `href` to a `fonts.googleapis.com` stylesheet after the page has loaded.

## Mechanism

Gate the **property setter** on the relevant prototypes:

- `src` on `HTMLImageElement` and `HTMLIFrameElement`
- `href` on `HTMLLinkElement`

A cross-origin resource whose URL matches a blocked provider in a denied category is **parked** (URL → `data-faz-src` / `data-faz-href`, no request — a `<link>`/``/`<iframe>` only loads with a live URL, which is never set when blocking) until consent, then restored by the standard `data-faz-src` / `data-faz-href` restore pass. Parked iframes are hidden and revealed on restore; tiles stay in place so map layout is undisturbed; a stylesheet loads only once its `href` is restored.

Tightly scoped and fast-pathed: same-origin, relative, `data:` and `blob:` resources bail immediately. The gate covers the `el.src = …` / `el.href = …` **property** assignment these libraries use; a URL committed via `setAttribute(…)` / `srcset` is out of this path's scope and stays covered by the existing markup / output-buffer blocking. `_fazImgCategory` tags the parked element with the category that actually triggers the block (multi-provider correctness).

## Verification

- jsdom unit suite (`tests/unit/js/img-iframe-src-gate.test.mjs`) — img/iframe/link gate: decision, fast-path bail-outs, faz-skip + whitelist escape hatches, per-category and per-service consent, protocol-relative, multi-provider category, Web Font Loader shape, `setAttribute` boundary.
- Real-browser repros: `tests/e2e/img-tile-blocking.mjs` (real Leaflet), `tests/e2e/font-link-blocking.mjs` (runtime Google Fonts link), `tests/e2e/specs/img-tile-blocking.spec.ts`.
- #163/#167 verified live on the reporter's Bricks sandbox; Web Font Loader gap reproduced and fixed locally. Confirmed the upstream (CookieYes 3.5.1) has the same runtime-injection gap and does not block Google Fonts at all.

## Summary by CodeRabbit

* **Bug Fixes**
  * Bloccate le map tile (Leaflet/OpenStreetMap e Bricks Map) quando la risorsa viene assegnata in modo runtime via `img`, con “parcheggio” fino al consenso e ripristino automatico dopo l’accettazione.
  * Bloccati anche gli embed lazy-load (iframe) con `src` impostato in runtime, nascosti durante il parcheggio e ripristinati post-consenso.
  * Gestite correttamente anche le stylesheet iniettate a runtime (es. Google Fonts/Web Font Loader): `href` viene parcheggiato fino al consenso.
  * Nessun impatto su risorse same-origin/relative e su `data:`/`blob:`.
* **Tests**
  * Nuovi test e2e di regressione per map tile, iframe e font link runtime pre/post-consenso.
* **Documentation**
  * Aggiornate note su changelog/README per la release **1.21.2**.
* **Chores**
  * Bump versione del plugin a **1.21.2**.


fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
## Code review — runtime resource-blocking gate (5 internal lenses, --full)

**Scope:** `main..HEAD` (9 commits). 21 raw candidates across L1 diff-local, L2 structural (opus), L3 CLAUDE.md, L4 comments, L6 security. No consent-bypass survives scrutiny; one genuine correctness bug, two minor fixes, the rest by-design / not-actionable.

### 🟠 Confirmed — worth fixing

**F1 · correctness · `frontend/js/script.js` (img/link/iframe restore passes)** — *convergent: L1 + L2*
The consent-restore passes do `el.src = imgSrc` / `el.href = fazHref` followed by an **unconditional** `removeAttribute("data-faz-src"/"data-faz-href")`. That assignment **re-enters the new property-setter gate**. If the URL matches a *second* provider whose category is still denied (or a per-service denial with no category denial), the gate re-parks the element (re-sets `data-faz-*`, never sets the native `src`/`href`) and the following `removeAttribute` strips the just-re-added marker → the element ends with **neither `src` nor `data-faz-src`** and is unrecoverable for the session (blank image/iframe, missing stylesheet). Narrow (needs a multi-provider/partial-consent URL) but real.
**Fix:** make the `removeAttribute` conditional on the native attribute actually being set (`if (el.getAttribute("src")) el.removeAttribute("data-faz-src")`), so a re-parked element stays parked and recoverable. **Not** `faz-skip` before the re-assign — that would force-load the resource even while another category is still denied (a consent-bypass).

**F2 · robustness · `_fazGateSrcSetter` / `_fazGateHrefSetter`** — *L1*
The descriptor guard checks `typeof desc.set === "function"` but **not `desc.get`**. If `desc.get` were `undefined`, the installed getter `nativeGet.call(this)` throws on every `.src`/`.href` read. Real browsers always expose the getter (no live impact), but the asymmetric guard is a one-line defensive fix.

**F3 · doc accuracy · CHANGELOG.md + test comment** — *L4*
The scope-boundary sentence says a URL committed via `setAttribute('src')` / `srcset` "stays covered by the existing markup / output-buffer blocking." For **runtime** `setAttribute`/`srcset` on main-document elements that is not true — the PHP `` handler only fires inside `<noscript>`, and the MutationObserver watches `script`/`iframe` only. Reword to "out of scope," not "covered elsewhere."

### 🟡 Minor / optional
- **F4** — the `href` gate parks **any** cross-origin `<link>` whose URL matches a provider, regardless of `rel` (preload/preconnect/dns-prefetch/icon). Harmless for hints, but consider gating on `rel=stylesheet`/`preload` to avoid touching non-stylesheet link types. *(L2)*
- **F5** — the comment "no per-resource provider scan on a normal page" is inaccurate for CDN-hosted media (cross-origin `cdn.*`/CloudFront/Photon images pay the provider scan on every `img.src`). Soften the comment. *(L2)*
- **F6** — the section JSDoc sits above `_fazImgShouldBlock` (a pure decision helper) while describing the prototype patching done by `_fazGateSrcSetter` (which has no JSDoc). Cosmetic. *(L4)*

### ✅ Reviewed, not actionable (by-design / not real)
- **Security (L6):** fail-open `try/catch` → native setter, `configurable:true` descriptor removable, `faz-skip` "attacker bypass," URL-parse-fail → don't-block — all assume an attacker already running JS on the page, who can bypass any client gate; the threat model is the **site's own** third-party scripts, against which these are correct (fail-open keeps the page working on error). TOCTOU on `_fazImgCategory` is not reachable — the two calls are synchronous with no yield.
- **getter returns empty on a parked element (L2):** deliberate — returning the parked URL was tried and reverted because it broke the real-Leaflet repro (`.src` falsely reading "loaded").
- **`_block` "dead switch" (L2):** defensive; always `true` server-side by design.

**Recommendation:** fix **F1** (real bug) + **F2** + **F3** before merge; F4–F6 optional.

fabiodalez-dev:
### ✅ Applied in `8dec7d4`

- **F1** — restore passes now guard `removeAttribute` on the native `src`/`href` actually being set (img/link/iframe): a re-parked resource stays parked + hidden and recoverable once the other category is also consented, instead of being stripped and bricked. Not `faz-skip` before the re-assign (that would force-load under a still-denied category — a consent bypass).
- **F2** — `_fazGateSrcSetter`/`_fazGateHrefSetter` now also require `typeof desc.get === "function"` before installing the accessor.
- **F3** — corrected the changelog/test scope wording: a runtime `setAttribute`/`srcset` assignment is out of scope (only server-rendered markup is covered by the output-buffer blocking).

Tests: jsdom 63/63 (new partial-consent multi-provider assertion for F1), real-Leaflet + Web Font Loader e2e pass, restore/consent specs (img-tile + frontend-consent + category-blocking) 19/19.

F4–F6 (rel-gating on the href gate, the CDN perf-comment wording, JSDoc placement) left as optional.

fabiodalez-dev:
Squash-merged into `main` as f25d1bb (conflict-resolved into the consolidated 1.22.0 release). Closing.

PR #177, opened 2026-07-06
Add soft cookie wall and popup center support

roboes:
# Add "Box (centered)" banner type and soft cookie wall overlay

## Description

FAZ Cookie Manager currently offers a box banner in a fixed corner only:

```php
<option value="box"><?php esc_html_e( 'Box (bottom corner)', 'faz-cookie-manager' ); ?></option>
```

Plugins like Complianz offer two features that are commonly requested for GDPR compliance UX:

1. A **centered modal box** - the banner appears in the middle of the screen rather than a corner, which draws more attention and is the dominant pattern on high-traffic European sites.
2. A **soft cookie wall** - a translucent overlay dims the page behind the banner, making clear to the visitor that a choice is required before interacting with the content, without fully blocking access.

Neither is currently possible in FAZ.

### Box (centered)

Adds a `popup` banner type that renders the box centered on screen via `transform: translate(-50%, -50%)`. The position dropdown is hidden when this type is selected (centering has no meaningful top/left variant). A new template entry is added to `template.json` for both supported template versions, cloned from the box template and extended with the centering rule.

### Soft cookie wall

Adds a `softCookieWall` boolean setting. When enabled, a full-screen `<div id="faz-cookie-wall">` is injected into `document.body` alongside the banner and removed automatically when the visitor makes a choice. The overlay styles are defined in the template CSS, setting it one level below the active banner container so that the banner remains fully interactive on top of the backdrop.

The toggle is hidden in the admin UI when the Classic type is selected, since Classic is a full-width bar where a backdrop makes less visual sense and the pushdown preference center expansion would conflict with it.

## Compatibility

| Banner type         | Centered box | Soft cookie wall |
|---------------------|:------------:|:----------------:|
| Box (bottom corner) | -            | ✅               |
| Box (centered)      | ✅           | ✅               |
| Full-width Banner   | -            | ✅               |
| Classic             | -            | ❌ hidden in UI  |

Existing banners are unaffected: `softCookieWall` defaults to `false`, and the new `popup` type is opt-in. The `popup` template entry in the JSON is only selected when the stored type is `popup`.

## Summary by CodeRabbit

* **Nuove funzionalità**
  * Aggiunto il banner **Box (centered)**, con posizionamento centrato.
  * Introdotta l’opzione **soft cookie wall**, che mostra uno sfondo semi-trasparente dietro al banner finché non viene fatta una scelta.

* **Miglioramenti**
  * Aggiornata la gestione delle posizioni per i banner popup e la relativa anteprima.
  * La nuova opzione è ora disponibile anche nelle impostazioni admin e supportata nelle configurazioni predefinite.

* **Bug Fixes**
  * Corretto il comportamento di visualizzazione e rimozione dell’overlay nella navigazione del banner.


fabiodalez-dev:
Thanks for this — a centered box and a soft cookie wall are both genuinely useful, and the core is well built: the setting is sanitized (`faz_sanitize_bool`) and escaped, the wall is removed on every consent path, the `_fazAddPositionClass` refactor preserves behaviour for the existing types, and the banner sits above the wall so Reject stays reachable.

I reviewed it in depth and there are a few things I'd need addressed before I can merge. Grouped by area:

**Process / housekeeping**
1. **Regenerate the minified JS.** The committed `frontend/js/script.min.js` isn't a clean `npm run build:min` of the final `script.js` (it has stray `;;` and the two new functions are positioned as if hand-patched). Please run `npm run build:min` and commit the regenerated file — the frontend loads the `.min`, so it must match source exactly.
2. **Drop the version bump.** `faz-cookie-manager.php` (`Version` + `Stable tag`) and `readme.txt` (`Stable tag`) were bumped to 1.23.0. Versioning happens at release time on my side, and a `Stable tag` change would mis-report to wp.org — please revert those to 1.22.0.
3. **Rebase on current `main`.** The branch predates the 1.22.0 release and a batch of geo-routing changes, so a few unrelated files show as reverted. A rebase will drop the noise.

**Correctness**
4. **Popup + Pushdown breaks the preference center.** The pushdown guards in `script.js` (e.g. `_fazGetPreferenceCenter`, lines ~1693/1712/1778) test `_fazGetPtype() === 'pushdown' && _fazGetType() !== 'box'`. `popup` shares box's structure (a separate `.faz-modal`, no embedded `.faz-preference-wrapper`) but isn't excluded, so a popup banner with `preferenceCenterType: 'pushdown'` routes into the pushdown-expand path and the Settings button opens nothing. Please exclude `popup` too, e.g. `['box','popup'].indexOf(_fazGetType()) === -1`.
5. **`softCookieWall` isn't reset when switching to Classic.** `updatePositionOptions()` hides `#faz-soft-cookie-wall-group` for classic but never unchecks/disables the input, and `syncFormToBannerData()` saves `.checked` unconditionally. So a banner that had the wall on keeps `softCookieWall: true` after switching to Classic — and since `#faz-cookie-wall` is in the classic template CSS, the wall still renders. Mirror what `preferenceCenterType` does (force-reset + `disabled`).
6. **Popup + Sidebar is a no-op.** There's no `popup-sidebar` template and `class-template.php` has no `popup → popup-sidebar` remap, but the Sidebar option stays enabled for the centered type, so it silently falls back to the modal. Please disable Sidebar for `popup` (or add the variant).

**UX / compliance**
7. **The wall blocks interaction — it isn't "soft".** `#faz-cookie-wall` has no `pointer-events: none`, so a fixed full-viewport div intercepts every click/tap: the page behind can be read and scrolled but nothing is clickable until a choice is made. That edges toward an interaction-blocking cookie wall (EDPB guidance disfavours conditioning access on consent). Please add `pointer-events: none` to make it a genuine soft dimmer (that also fixes the keyboard/mouse inconsistency, since keyboard users can currently Tab past it).
8. **Fix the help copy.** "Dims the page behind the banner. Visitors can still scroll but see the content obscured…" reads as pure dimming; once #7 is in it's accurate, otherwise it should say interaction is blocked.
9. **Add a short legal-risk note by the toggle.** Since this is a compliance-oriented plugin, a one-line caution that cookie-wall patterns carry EU consent-validity risk (enable after legal review) would keep an operator from assuming it's automatically compliant.

Points 1–3 and 7 are the important ones. Happy to help if any of these are unclear — and thanks again for taking the time to build this.

fabiodalez-dev:
One refinement on point 5 (CodeRabbit flagged the same thing): please guard `softCookieWall` **server-side** too, not just by resetting the checkbox in the editor. A direct REST write to `PUT /wp-json/faz/v1/banners/{id}` can carry `softCookieWall: true` with `type: 'classic'` and bypass the admin JS entirely, so a client-side reset alone leaves the wall renderable on a Classic banner.

Force it to `false` for classic in `class-banner.php` — either in the sanitize path (the `faz_sanitize_bool` switch is fine as-is; add a post-sanitize rule) or in `apply_runtime_layout_compatibility()`, so the invariant "Classic never has a soft wall" holds regardless of how the banner is saved. Keep the checkbox reset in `banner.js` for the editor UX, but the server guard is what makes it robust.

roboes:
Done - addressed all points. Let me know if anything needs a tweak.

fabiodalez-dev:
A note on the compliance side of the soft cookie wall, since it shaped how I landed it.

A wall that actually blocks the page — no scroll, no interaction — until the visitor makes a consent choice conditions access on consent. Under EDPB Guidelines 05/2020 on consent (§39), consent obtained that way generally isn't "freely given," so a blocking cookie wall is a real GDPR risk in the EU. That runs against what this plugin is built to do: it defaults to the opt-in floor and stays clear of dark patterns.

So I kept the overlay but made it a **visual cue only**. It's rendered with `pointer-events: none` and `aria-hidden="true"`: it dims the page to draw attention to the banner, but never blocks reading, scrolling, or clicking. The admin help text and a short legal caution now say exactly that, it's disabled for the Classic layout, and it's off by default.

If a site owner in a jurisdiction where blocking walls are acceptable wants a true blocker, that's a follow-up I could add later as an explicit, separately-gated option with its own warning — but the default here stays compliance-safe.

PR #185, opened 2026-07-12
Fix: guard _fazInit against double-execution under Cloudflare Rocket Loader

roboes:
# Fix: guard _fazInit against double-execution under Cloudflare Rocket Loader

## Description

When Cloudflare Rocket Loader is enabled, `script.min.js` is executed twice per page load. Rocket Loader processes the script through its own deferred execution queue even when `data-cfasync="false"` is present. The second execution calls `_fazDomReady` $\rightarrow$ `_fazInit` $\rightarrow$ `_fazRenderBanner`, inserting a second copy of the banner into the DOM.

`_fazGetBanner()` uses `document.querySelector`, which returns the first match. Clicking Reject hides banner 1 while banner 2 (inserted by the second `_fazInit` run) stays visible. The consent cookie is written correctly (the banner does not reappear on reload), but the banner never disappears on the current page, breaking the UX.

### Root Cause

There is no idempotency guard on `_fazInit`. The fix requires only that it runs at most once per page load.

## Solution

Added a `_fazInitDone` flag on the persistent `window.fazcookie` object, which survives re-execution because it is initialized as `window.fazcookie = window.fazcookie || {}`. The `_fazDomReady` callback checks this flag and short-circuits on the second execution before `_fazInit` can render a duplicate banner.

## Summary by CodeRabbit

* **Correzioni**
  * Evitata la doppia inizializzazione del banner dei cookie quando gli script vengono caricati più volte.
  * Migliorata la compatibilità con strumenti di ottimizzazione e caricamento differito degli script.


fabiodalez-dev (approved):
Approving — this is a correct, well-scoped fix. I verified:

- **Right global**: `window.fazcookie` is the plugin's actual persistent object (`const ref = window.fazcookie` at script.js:34), not an invented one, so the flag genuinely survives the second execution.
- **No race**: the flag is set synchronously before `await _fazInit()`, so the second run sees `_fazInitDone === true` before the first run yields at the await. JS is single-threaded, so this holds.
- **No legitimate re-init blocked**: `_fazInit` has a single call site (this bootstrap). The language swap happens *inside* `_fazInit`, not by re-invoking it, so the guard can't break it.
- **`script.min.js` is in sync**: regenerated with the pinned terser, byte-identical to `npm run build:min`, and `_fazInitDone` is present in the shipped minified file.

One follow-up thought (not a blocker for this PR): the guard stops the second `_fazInit`, which fixes the reported double-banner. If Rocket Loader also re-runs top-level blocking-patch code, that's a separate, lower-risk surface — most of it is idempotent and the internal registration helpers (e.g. `_fazRegisterShortcodeTriggers`) already carry their own guards, so it isn't a concern for the reported symptom.

I'll add a jsdom regression test (double DomReady firing → `_fazInit` runs once) and fold this into the next patch release. Thanks for the clean fix — and for regenerating the min file this time.

roboes:
Hi @fabiodalez-dev, thanks for approving and merging this PR! Is there an estimated timeline for the next release containing this update?

fabiodalez-dev:
Hi! Maybe tomorrow!

PR #190, opened 2026-07-22
Feature: Add placeholder blocking for Smash Balloon Instagram Feed and Elementor Video widget

roboes:
# Feature: Add placeholder blocking for Smash Balloon Instagram Feed and Elementor Video widget

## Description

Two third-party embed types were rendering unblocked before consent, bypassing the existing Google Maps-style placeholder mechanism entirely.

### 1. Smash Balloon Instagram Feed

[Smash Balloon](https://wordpress.org/plugins/instagram-feed/) renders as `<div id="sb_instagram" class="sbi ...">`. The existing social embed detector (`process_social_embeds()`) only matched embeds by class (`fb-page`, `instagram-media`, `twitter-tweet`, …) and was only ever invoked from `filter_content_blocking()`, hooked to `the_content` / `widget_text` / `widget_block_content`.

A Smash Balloon widget placed via a page builder outside the main post content (header, footer, theme-builder template) never passed through those filters and was never scanned at all - regardless of id/class matching.

### 2. Elementor Video Widget

Elementor's native Video widget renders an empty `<div class="elementor-video"></div>` server-side. The real embed URL only exists inside `data-settings="{&quot;youtube_url&quot;:...}"` JSON on the ancestor widget wrapper - there is no `<iframe src="...">` in the initial HTML.

The plugin's generic iframe blocker (`process_iframe_tag()`) matches on a `src` attribute, so it never had anything to catch for this widget type.

## Root Cause

1. `process_social_embeds()` was id-blind (class-only) and wired to a narrower set of content filters than the full-page output buffer, so any social/feed widget rendered outside post content was invisible to it.
2. No detector existed for page builders that defer real embed construction to client-side JS reading a JSON attribute instead of emitting a `src`-bearing element server-side.

### `process_social_embeds()` - id match + broader hook

- Extended with an id-anchored match for `sb_instagram`, reusing the same block/placeholder/per-service-consent logic as the existing class-anchored loop.
- Added a matching Instagram-branded icon/color entry (`smash-balloon-instagram`) in `Placeholder_Builder`.
- Wired into `process_output_buffer()` (the full-page buffer), not just `filter_content_blocking()`, so social/feed embeds are caught regardless of where on the page they render.

### `process_elementor_video_widgets()` - new detector

- Detects `.elementor-widget-video` wrappers.
- Decodes their `data-settings` JSON and resolves the embed URL (`youtube_url` / `vimeo_url` / `dailymotion_url`) through the existing `Placeholder_Builder::detect_service_from_url()`.
- Gates the whole wrapper the same way as other blocked embeds.
- Wired into both `filter_content_blocking()` and `process_output_buffer()`.

### `Placeholder_Builder::build_social()` — video-aware card sizing

- Extended to render a 16:9 `faz-placeholder--video` card (brand color + service name) when the resolved service is a video platform, instead of the compact 120px social-icon card.
- Prevents a blocked video widget from collapsing to a tiny box and causing a layout shift.

---

## Files Changed

| File | Change |
|---|---|
| `process_social_embeds()` | Added id-based `sb_instagram` match; hooked into output buffer |
| `process_elementor_video_widgets()` | New function; hooked into both content filters and output buffer |
| `Placeholder_Builder` | Added `smash-balloon-instagram` entry; video-aware card sizing in `build_social()` |

## Summary by CodeRabbit

* **Nuove funzionalità**
  * Aggiunto il supporto per bloccare e sostituire con un segnaposto i widget video di Elementor quando gli script non sono autorizzati.
  * Aggiunto il supporto per i feed Instagram di Smash Balloon, inclusa la gestione del consenso dedicato.
  * Migliorata la visualizzazione dei segnaposto per i contenuti video, con layout e proporzioni corrette.

* **Correzioni**
  * La gestione dei contenuti bloccati ora si applica all’intera pagina, inclusi widget ed elementi incorporati al di fuori dei post.

### DIFF
--- a/frontend/class-frontend.php
+++ b/frontend/class-frontend.php
@@ -2666,6 +2666,18 @@ class Frontend {
 			$html = strtr( $html, $noscript_stash );
 		}
 
+		// 7. Hide social/widget embed containers that depend on blocked scripts
+		// (Facebook/Instagram/Twitter/X, Smash Balloon Instagram Feed, Elementor
+		// video widgets). filter_content_blocking() already runs these same two
+		// detectors for the_content/widget_text/widget_block_content, but an
+		// embed placed OUTSIDE post content - a theme-builder header/footer, a
+		// global (non-widget-area) template part, or any markup a page builder
+		// renders directly into the page - never passes through those narrower
+		// filters and only ever reaches the page here, in the full-page buffer.
+		// Without this, such an embed is never detected at all.
+		$html = $this->process_social_embeds( $html, $blocked_categories );
+		$html = $this->process_elementor_video_widgets( $html, $blocked_categories );
+
 		return $html;
 	}
 
@@ -6085,6 +6097,11 @@ class Frontend {
 		// Hide social embed containers that depend on blocked scripts.
 		$content = $this->process_social_embeds( $content, $blocked_categories );
 
+		// Hide Elementor video widgets (they render an EMPTY .elementor-video
+		// wrapper server-side and build the real iframe client-side from
+		// data-settings, so the generic <iframe> blocker above never sees one).
+		$content = $this->process_elementor_video_widgets( $content, $blocked_categories );
+
 		return $content;
 	}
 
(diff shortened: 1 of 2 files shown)


fabiodalez-dev:
Thanks for this — the Elementor Video detector is a genuinely good catch. An empty `<div>` whose URL lives only in the `data-settings` JSON is invisible to the iframe blocker, so nothing was gating those embeds at all.

I pushed one commit to your branch (`8b3d20b`) rather than sending you back and forth over a five-line change. CodeRabbit was right about the Smash Balloon pass: it assigned `preg_replace_callback()` straight to `$content` with no null check. `filter_content_blocking()` is hooked to `the_content` / `widget_text` / `widget_block_content`, and a `null` return from a content filter doesn't pass the original through — it renders as an empty string, so a PCRE failure there would silently blank the whole post body. Low probability (it needs `pcre.backtrack_limit`), but the blast radius made it worth closing.

You had already got this right in `process_elementor_video_widgets()` with `null !== $result ? $result : $content`, so it was just the one block. I used the shape the same function already uses for its scripts and iframes passes, which also logs `preg_last_error()`, and fixed the indentation on the callback's `return` while I was there.

The other half of CodeRabbit's note — applying the same guard to the pre-existing class-based social loop just above yours — is on code that predates this PR, so I left it out of your diff and will handle it separately.

Everything else reads well: the id-anchored match for `sb_instagram*`, keeping `smash-balloon-instagram` as its own service id instead of folding it into the native oEmbed entry, the per-service consent override mirroring the class-based path, and decoding the HTML entities before `json_decode` on `data-settings`.

roboes:
Thanks!

PR #189, opened 2026-07-22
Competitor-parity features: third-country consent, age gate, anti-adblock, setup wizard, A/B testing

fabiodalez-dev:
Five competitor-parity features, each opt-in and default OFF so existing installs are unchanged until enabled. Built to release quality (full escaping/nonce/sanitization, i18n, backward-compat, tests) after reviewing what Complianz and Real Cookie Banner offer that FAZ lacked.

## Features

| Feature | What it does |
|---|---|
| **Third-country consent (Schrems II)** | Per-cookie "transfers personal data to insecure third countries" flag → a disclosure in the preference center and an "International data transfers" section in the generated Cookie Policy. Labels resolve in the banner's active language, not the site locale. |
| **Youth / age gate (GDPR Art. 8)** | Optional age-confirmation checkbox (threshold configurable 13–16, default 16) that gates only the Accept path; Reject/withdraw is never gated; the age affirmation is recorded alongside the consent. |
| **Anti-adblock resilience** | Re-shows the legally-required banner if an ad-block cosmetic filter list hides it — a single deferred re-assert, no loop, never a cookie wall, never re-enables trackers. |
| **Setup wizard** | Guided first-run flow (jurisdiction → optional scan → finish) that writes compliant defaults for the chosen law; a dismissible "complete setup" card on the dashboard. |
| **A/B testing of banner variants** | Persistent random split across 2+ banners with per-variant accept-rate reporting from the consent log, shown on the dashboard. Disabled under Cache Compatibility Mode (documented) to keep cached HTML visitor-invariant. |

## Compliance framing

- Every A/B variant stays independently compliant (equal-weight buttons, no pre-check); the tool only chooses among existing compliant banner rows.
- The age gate is not a substitute for parental-consent mechanisms (stated in the help text); the threshold is admin-configurable because the digital age of consent varies by member state.
- The wizard writes opt-in + no-pre-check + equal-weight + Garante-capped-expiry defaults for GDPR-family laws and never nudges toward a less-protective setting.
- Anti-adblock is resilience for a required notice, not circumvention of privacy tools.

## Design notes

- A/B server-side split conflicts with full-page caching, so it is disabled under Cache Compatibility Mode; a cache-safe client-side split is left as a documented future extension.
- With the advanced `faz_geo_ruleset_runtime` flag on, jurisdiction routing can override the A/B variant (compliance beats the experiment) — documented in the method docblock.

## Verification

[...]

PR #194, opened 2026-07-26
test(e2e): fix the blocker-template rule-count test (rotted by PR #170) and its state pollution

fabiodalez-dev:
The v170-deep-flows blocker-template test hardcoded a +3 row expectation from when the Google Analytics template shipped 3 patterns; #170 grew it to 52, and the test has been failing since. It also polluted the shared test site on every run (the template click saves the rules and nothing cleaned them up) and deadlocked on an empty baseline (visibility wait on a zero-height tbody).

The test now reads the expected delta from the shipped template JSON, snapshots and restores `script_blocking.custom_rules` via WP-CLI around the click, and waits on the settings GET response instead of tbody visibility.

Verified locally against the deployed test site: the test passes and the option is byte-identical before/after.

## Summary by CodeRabbit

* **Test**
  * Migliorata l’affidabilità dei test end-to-end relativi ai template di blocco.
  * Il numero di regole personalizzate atteso ora viene calcolato dinamicamente in base al contenuto del template selezionato.
  * Rafforzata la verifica delle regole iniziali e la sincronizzazione durante il caricamento dei dati.
  * Migliorato il ripristino dello stato a fine test, con gestione più sicura degli scenari di recupero.

### DIFF
--- a/tests/e2e/specs/v170-deep-flows.spec.ts
+++ b/tests/e2e/specs/v170-deep-flows.spec.ts
@@ -1,5 +1,6 @@
 import { execFileSync } from 'node:child_process';
 import { existsSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
 import { dirname, join } from 'node:path';
 import type { Page } from '@playwright/test';
 import { expect, test } from '../fixtures/wp-fixture';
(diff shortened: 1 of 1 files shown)


fabiodalez-dev:
Superseded by #189, which already fixes this test — and fixes it better.

I diagnosed the hardcoded `+3` on the POPIA branch (off main, where the test is still broken) without checking whether an open PR had already dealt with it. #189 does: it reads the expected patterns from `/faz/v1/blocker-templates` and resets `custom_rules` through the REST API, which is both closer to what the admin actually does and free of the WP-CLI dependency my version introduced.

Closing as duplicate. The finding stands, the fix is already in flight.

PR #197, opened 2026-07-29
Cross-host cookie scan, verified wizard persistence, accurate POPIA framing

fabiodalez-dev:
Follow-up to the shared scan engine on `main`, plus a legal-framing correction for POPIA. Four commits, each independently reviewable.

## Scanner: cross-host paths and an honest failure

WordPress can serve `wp-admin` and the public home URL on different hostnames — www/apex, a reverse proxy, a mapped admin domain. The browser cannot read a cross-origin iframe, so those pages were skipped and the scan reported nothing without explaining why.

Discovered paths are now retried through the exact admin origin, which is same-origin by construction and usually serves the same front controller. Only `pathname + search + hash` are carried over, so a rebase can never reach a third-party host; when the alias does not serve the same content, the canonical redirect leaves the iframe cross-origin and is still recorded as `iframeInaccessible`.

Separately: if every page failed and nothing at all was observed, the run used to continue into the server-side enrichment pass and import its result — recreating the "one or two cookies found" answer the browser engine exists to replace. That case now rejects with an actionable message. The guard requires zero cookies **and** zero scripts as well as an issue on every page, and `originRebased` deliberately does not feed `totalIssues`, so a fully rebased, fully successful scan cannot trip it.

Same-origin detection now compares `URL.origin` directly. The previous helper stripped a leading `www.` before comparing, treating www and apex as one origin — which the browser does not, so the read failed later and was blamed on framing rather than on the host mismatch.

## Wizard: verify persistence before reporting success

`finish()` wrote the settings and the GCM option, then reported success unconditionally. `update_option()` returning `false` is ambiguous — it means "unchanged" as often as "write failed" — so a genuinely failed write produced a wizard that said it was done while the site kept its old configuration.

Both writes are now confirmed by a fresh, sanitised read-back, and a mismatch returns a `WP_Error`. The read-back is trustworthy because `Settings::update()` clears the static settings cache, so `get()` genuinely re-reads the option; the comparison is exact because both sides pass through the same sanitiser against the same defaults, which builds its result by iterating the defaults and therefore yields a deterministic key order.

Proven by fault injection rather than inspection: the runtime unit suite makes a settings write silently no-op and asserts `finish()` reports failure.

## POPIA: describe the preset, not a claim about the Act

The wizard copy, the policy templates and the test commentary described POPIA as "an opt-in regime" in which consent is the basis for every non-essential technology. That overstates it. Section 11(1)(a) consent is what this preset chooses — deliberately, as the conservative option — but section 11(1)(b)-(f) permits other lawful justifications, and a site relying on one of those must assess and document it separately. Telling an administrator that POPIA simply requires consent invites a policy that misstates its own legal basis.

Corrected in the wizard law card, the Cookie Policy in all eight bundled languages, the gettext catalogue, the compliance-suite commentary and the changelogs. Behaviour is unchanged: templates keep their ten sections and an identical placeholder multiset across languages, and the POPIA assertions are untouched — nothing non-essential runs before the visitor acts, no category arrives pre-ticked, refusing is no harder than accepting.

## i18n

Five new strings, one dropped. `msgmerge` had fuzzy-matched both save-failure messages onto the same generic *"Impossibile salvare la configurazione"*, erasing the one thing those errors exist to convey — which write failed; both were rewritten. Italian 1390 → 1395 translated, zero fuzzy. All six `.mo` recompiled and byte-matching a fresh `msgfmt` of their `.po`.

## Verification

- Unit suites: **55/55** (including the new `js/scan-engine.test.mjs`)
- Compliance suite: **122/122**, with GCM enabled — section 8 GCM 9/9, section 15 POPIA 9/9
- Cookie Policy gettext catalogue: byte-identical to a fresh regeneration, so the msgids still match the shipped `en.md`
- POPIA templates: 10 H2 sections and an identical placeholder multiset across all eight languages
- PHP/JS syntax and `git diff --check`: clean
- Full E2E suite: running against the branch; result posted below when it completes

## Summary by CodeRabbit

- **Nuove funzionalità**
  - La scansione dei cookie nel wizard usa il motore browser condiviso con la pagina Cookies, con gestione degli host differenti e diagnostica migliorata.
  - Il completamento del wizard verifica ora che banner, impostazioni e Google Consent Mode siano stati salvati correttamente.

- **Correzioni**
  - Evitata l’importazione di risultati non verificabili quando nessuna pagina può essere osservata.

- **Documentazione**
  - Aggiornati preset e template Cookie Policy POPIA con riferimenti più precisi al consenso e alle basi giuridiche alternative previste dalla normativa.


fabiodalez-dev:
## Verification — all green

| Suite | Result |
|---|---|
| Full Playwright E2E | **1013 passed, 7 skipped, 0 failed, 0 flaky** (48m) |
| Unit suites | **55/55**, including the new `js/scan-engine.test.mjs` |
| Compliance | **122/122** — section 8 GCM 9/9, section 15 POPIA 9/9 |
| Cookie Policy gettext catalogue | byte-identical to a fresh regeneration |
| `faz-cookie-manager.pot` | 1419 msgids, matches a fresh `make-pot` |
| Bundled `.mo` | all six byte-match `msgfmt` of their `.po` |
| Plugin Check (wp.org shape) + CodeQL | green |

Same count as the pre-change baseline on `main` (1013 / 7 / 0), so nothing regressed.

PR #199, opened 2026-07-31
Plan the legal documents generator, and pin the Cookie Policy render before touching it

fabiodalez-dev:
This branch carries two commits: the implementation plan for the legal documents generator, and the regression test that pins the Cookie Policy render before I touch it.

## Why the plan generalises the engine in place

The obvious move would be a second module sitting next to the Cookie Policy generator. I checked that assumption against the code and rejected it.

`admin/modules/cookie-policy-generator/` is roughly 3,930 lines of PHP that is already shaped like a general document generator with one document type hardcoded into it:

- `Generator` is pure-static and pure-function. `resolve_template_path()`, `substitute()`, `markdown_to_html()` and `policy_version_hash()` contain zero cookie-specific logic. The cookie-specific surface is exactly three constants (`JURISDICTIONS`, `HTML_TOKENS`, `NATIVE_LANG`) plus one hardcoded directory in `templates_dir()`.
- `Section_Overrides` already operates on a `$settings` array passed in, so it is document-agnostic today. Giving each document its own settings option gains the document dimension for free, with no re-keying of existing data.
- `Template_Translations` is parameterised by everything except `CATALOG_FILE` and the English source path.
- `Renderer` is the only genuinely coupled class, but its pipeline — resolve language, resolve jurisdiction, load scaffold, gettext, overrides, substitute, markdown, disclaimer, kses, hash — is exactly the pipeline every legal document needs.

A parallel module would fork the markdown pipeline, the gettext catalogue machinery, the section-override editor and the sanitisation patterns. All four took multiple point releases (1.16.1, 1.16.2, 1.23.1, 1.25.0) to harden against real user reports: the empty-label lines, the curly-quote shortcode attributes, the JSON-in-category-names leak, the i18n placeholder-parity guards. Forking that code means fixing every future bug twice. Threading a `Document_Config` parameter through four classes costs far less than maintaining a second pipeline.

The plan also corrects a few working assumptions I had going in. `missing_required_settings()` gates POPIA only, so per-document and per-jurisdiction refusal logic is real architectural work rather than template text. The `faz/cookie-policy` Gutenberg block renders the *legacy* canned policy, not the engine, so new documents need their own block while the old one stays untouched for backward compatibility. A consent-invalidation REST endpoint already exists, so the version-to-revision work needs a trigger surface and an acknowledgment ledger rather than a new bump mechanism.

## Why the golden test lands first

The refactor described above rewrites the inside of a renderer whose output is a legal document on roughly a thousand live sites. Output that changes silently is the failure mode that matters here, and it is not one that ordinary unit tests catch — every individual function can keep behaving correctly while the assembled document drifts.

So this commit pins the current output byte-for-byte, before any of that work starts. It is the backward-compatibility gate for the refactor that follows. If a change moves one byte of any fixture, the suite fails, and the correct response is to fix the code rather than to regenerate the goldens.

## What the 16 fixtures cover

Each fixture pins one branch of the pipeline:

| Fixture | Branch pinned |
|---|---|
| `gdpr-empty-install` | no cookies, no company data |
| `gdpr-full-en` / `gdpr-full-it` | fully populated, native and translated |
| `gdpr-show-title` | title-rendering flag |
| `gdpr-empty-label-lines` | the empty-label regression from 1.16.x |
| `gdpr-transfers` | third-country transfer disclosures |
| `gdpr-section-override` | a section override applied |
| `gdpr-section-override-anchor-drift` | override whose anchor no longer matches |
| `gdpr-gettext-merge` | the gettext catalogue path |
| `ccpa-california`, `lgpd-brazil`, `popia-complete` | the other three jurisdictions |
| `popia-incomplete-refuses` | refusal output when required data is missing |
| `popia-incomplete-visitor` | what a visitor sees in that same state (empty) |
| `unknown-jurisdiction-att-ignored` | unrecognised jurisdiction falls back safely |
| `tables-absent` | cookie tables missing from the database |

Two of these encode deliberate non-obvious behaviour worth calling out. `Section_Overrides::apply()` returns the *original* string when nothing applies, rather than re-joining the parts, because the effective scaffold is hashed into `data-faz-policy-version` and re-joining would bump that hash on every install — `gdpr-section-override-anchor-drift` pins that. And `popia-incomplete-visitor` is an empty file on purpose: a visitor gets nothing at all, while an administrator gets the refusal notice.

The harness resets its state between fixtures. Without that reset the second fixture in a run would reuse the first one's cookie inventory and every golden after the first would be quietly wrong.

## Why wp_kses is stubbed as identity

`wp_kses()` is stubbed to return its input unchanged, so the golden files record the pre-sanitisation HTML. Reimplementing kses in the harness would pin WordPress's behaviour rather than the plugin's, which is not what this suite is for.

That leaves a real gap, because the allowlist matters here. `render()` does not use plain `wp_kses_post()` — it passes a custom allowlist that adds `role` and `aria-level` on top of the `post` context. Without those, `wp_kses_post()` strips `aria-level` and leaves category names as headings with no level, which is an accessibility regression that the goldens alone would not catch.

So the allowlist is captured from the last `wp_kses()` call and asserted separately: the suite checks that `render()` passes a custom allowlist at all, and that the allowlist still carries those accessibility attributes. Identity stub for the bytes, explicit assertion for the sanitisation contract.

## Verification

I confirmed the suite actually fails rather than passing vacuously. Shortening the version-hash prefix by a single character fails all sixteen goldens, and changing one byte of one fixture's hash fails that fixture — `FAILED — 1 of 65 checks failed`. A test that has never been observed failing is not a gate.

`npm run test:unit` goes from 56 suites to 57, all passing; the new suite reports 65/65 checks. The PHP syntax sweep is clean. No plugin code is touched by this branch — it adds a plan document, one test file and its fixtures.

## Summary by CodeRabbit

* **Documentazione**
  * Aggiunto un piano per evolvere il generatore di Cookie Policy in un sistema per più documenti legali.
  * Previsto il supporto per Privacy Policy, Termini e condizioni e Disclaimer.
  * Documentati template, shortcode, blocchi, API, impostazioni amministrative e collegamenti nel footer.
  * Inclusi requisiti per dati obbligatori, fallback linguistici, compatibilità, test, versionamento del consenso e rilascio.


fabiodalez-dev:
Thanks — seven of the eight are fixed in cc4d2ad. Two of them would have caused real defects, so they were worth catching:

- The documented pipeline order was wrong. `register_version_meta()` runs at `class-renderer.php:212`, before `wp_kses()` at 235, so the hash covers the pre-sanitisation HTML. Implementing the order as written would have moved `data-faz-policy-version` on every install and broken the byte-for-byte guarantee the golden test enforces.
- The `data_builder` callable point is correct and I had missed it. PHP resolves visibility for string/array callables at the call site, so a callable registered in `Document_Registry` could not reach `Renderer`'s private static helpers. The plan now specifies dispatch inside `Renderer`, plus a test that executes each registered builder so a visibility fatal surfaces in the suite.

I also took the `invalidate_consents()` point further than suggested. You are right that a Terms change must not re-open the cookie banner, so the registry now carries an `affects_cookie_consent` flag per document; when false the re-consent action is not rendered at all rather than rendered-and-disabled.

One I did not apply: the i18n resync language list. The plan lists `cs/de/fr/hr/it/nl`, which is exactly the set of `.po` catalogues in `languages/` — those are the locales the plugin's interface is translated into. `Generator::LANGUAGES` (`en it fr de es pt-BR bg cs`) is a different set: the languages a policy *scaffold* ships in. They overlap but neither contains the other, and merging them would run msgmerge against files that do not exist. The list was already right — but the document never said which of the two sets it meant, which is presumably what made the comparison look wrong, so both are now named explicitly.

PR #201, opened 2026-07-31
feat: detect Cookie Policy version changes and let the operator decide on re-consent

fabiodalez-dev:
## What this does

The renderer has stamped every Cookie Policy with a content hash (`data-faz-policy-version`) since 1.16.0, but nothing ever recorded which hash the administrator had actually *seen*. So the plugin could tell that a policy had changed and still had no way to raise the one question that matters: is this change material enough to require fresh consent?

This PR records the acknowledged hash, and when the current one differs it shows a notice on the Cookie Policy admin page offering two explicit paths:

- **"This was a minor change — keep existing consents"** — acknowledges only. Nothing else moves.
- **"Material change — ask visitors to consent again"** — confirms the consequence, calls the **existing** `POST faz/v1/settings/invalidate-consents`, then acknowledges.
- **"Decide later"** — hides the notice for that page load, persists nothing.

Nothing happens automatically, and the plugin's own upgrade never prompts.

## Why it is shaped this way

**The decision is not the plugin's to make.** Re-prompting every visitor has a real cost (consent fatigue, lost analytics, support tickets); not re-prompting after a substantive rewrite is a compliance failure. Only the operator knows which just happened, so I made the ledger do the least it can — remember a hash, report whether it still matches — and put every consequence behind an explicit click.

**The minor button is the `faz-btn-primary`, deliberately.** Most version moves are a template tweak or a shipped translation update. A notice that visually pushes towards invalidation would train operators to click the loud button without reading it, which is how re-prompt fatigue starts. I'd rather a reviewer reject a styling swap here than have this become a nag.

**`current_hash()` pins `$_SERVER['REQUEST_URI']` to `/` and restores it in a `finally`.** `{{COOKIE_POLICY_URL}}` is built from the request path (`Renderer::current_url()`) and *is* part of the hashed data. Without the pin, the hash computed from `/wp-admin/admin.php?page=…` would never equal the one stored a moment earlier and the notice would fire spuriously forever. The pinned value only needs to be self-consistent — it will legitimately differ from the hash rendered on the public policy page, whose own path is part of its data. I added a unit assertion that renders *without* the pin and asserts the hash differs, so the pin can't quietly become dead code.

**The hash is read back off the markup, not recomputed.** `data-faz-policy-version` is the value the golden-render suite pins; a second computation path would be free to drift from it.

**Seeding is silent, and never happens when nothing renders.** The first evaluation on a configured install adopts the current hash without prompting — greeting existing installs with a re-consent prompt purely because they updated would be exactly the wrong first impression. And when the policy refuses to render at all (a POPIA install whose mandatory operator fields are still blank, which emits the refusal notice and no hash), the option is *not* written: seeding an empty entry would make the operator's first real configuration read as a change and fire a bogus prompt.

**The client sends the hash it saw, not a server recompute.** If the policy changes again between page load and click, acknowledging the older hash is the honest outcome — the next page load compares against it, finds a difference, and asks again. Recomputing server-side would silently swallow that second change.

**No second bump mechanism.** `invalidate_consents()` already exists at `admin/modules/settings/api/class-api.php` and already handles the awkward bit (it re-reads the persisted revision because `Settings::sanitize_option('consent_revision')` caps at 999999 and refuses to lower). The new endpoint only records a hash. One caveat I hit: `cookie-policy.js`'s local `api()` helper hardcodes the `cookie-policy/` REST prefix, so `api('POST','settings/invalidate-consents')` would hit `faz/v1/cookie-policy/settings/invalidate-consents` and 404 — the material path uses `window.FAZ.post` directly, with a comment saying why.

**"Decide later" is client-side only.** No dismissed-flag option, no `is-dismissible` persistence. A change the operator has not judged should keep asking.

## Backward compatibility

- **No frontend changes at all.** The consent-revision plumbing already runs end to end (`class-frontend.php` → `policyRevision`/`_consentRevision` → `script.js`). No `build:min` needed.
- **No migration, no activator change.** Seeding is lazy: the option appears the first time someone opens the Cookie Policy admin page on a configured install.
- **The new option is `faz_legal_doc_acknowledged`, keyed by document slug** even though only `cookie-policy` exists today, so further generated legal documents slot in without a migration. Stored with `autoload = false` — it is read on exactly one admin screen.
- **Rendered policy output is byte-identical.** The ledger only reads what the renderer already produces.
- `uninstall.php` deletes the new option in the explicit list, under the existing `remove_data_on_uninstall` gate.
- Statuses `seeded` / `unchanged` / `unavailable` render nothing at all, so an install that never changes its policy sees no new markup.

One cost worth naming: `evaluate()` renders the policy once per Cookie Policy admin page load. That is one extra inventory query on a single admin screen, and it is what makes the comparison trustworthy rather than approximate.

## What I tested

- `npm run test:unit` — **57 suites passed, 0 failed**, including the new `tests/unit/test-version-ledger.php` at **31/31 checks**. It covers: silent seeding (and that a second evaluation is quiet), the stored entry's shape and `autoload=false`, change detection after editing the controller name and settling once acknowledged, rejection of `abcdef` / `ABCDEF.ABCDEF` / `abcdef.abcdef1` / `<script>` / a non-string with the option left untouched, the POPIA no-render case leaving the option unwritten, REQUEST_URI independence (plus the unpinned control described above, and that `REQUEST_URI` is restored), and hash stability when the calendar advances a day.
- PHP syntax sweep across the plugin — clean.
- `npx eslint admin/assets/js/pages/cookie-policy.js` — clean.
- `tsc --noEmit --strict` on the new spec — clean.
- I did **not** run the Playwright E2E suite: it drives the shared WordPress test site and parallel runs corrupt each other's fixtures. `tests/e2e/specs/policy-revision-link.spec.ts` is committed and covers the visitor-visible half — a consented browser sees the banner again after "Material change", does not after "Minor change", "Decide later" brings the notice back on the next load, and the settings snapshot is restored in `afterAll`. It needs to run in the separate E2E pass.

## Summary by CodeRabbit

* **Nuove funzionalità**
  * Aggiunta la notifica per le modifiche alla Cookie Policy.
  * È possibile mantenere i consensi, richiederli nuovamente o rimandare la decisione.
  * Le modifiche riconosciute vengono registrate automaticamente.

* **Correzioni**
  * Gli errori nelle operazioni vengono mostrati chiaramente e la notifica scompare solo dopo il completamento corretto.

* **Test**
  * Aggiunti test end-to-end per verificare i flussi relativi a modifiche minori, sostanziali e decisioni rimandate.

* **Documentazione**
  * Aggiunto il piano per estendere il generatore anche ad altri documenti legali.


fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai full review

PR #202, opened 2026-07-31
Collect what installed plugins declare about their own data processing

fabiodalez-dev:
## What this does

Adds `Content_Collector`, a data layer that reads every privacy-policy declaration the installed plugins register with WordPress and snapshots them into a plugin option.

WordPress has had the mechanism since 4.9.6. A plugin calls `wp_add_privacy_policy_content( $plugin_name, $policy_text )` to describe, in its own words, what it processes; `WP_Privacy_Policy_Content::get_suggested_policy_text()` hands back every registered contribution. FAZ is already a **producer** of one such block — `CLI::register_privacy_hooks()` registers the consent-log disclosure. This makes it a **consumer** as well, which is the symmetric and obvious move.

Nothing renders yet. This PR is the collection, diff and override layer plus its tests; the renderer, the admin screen and the `Document_Registry` wiring are follow-ups (listed at the bottom).

## Why aggregate, instead of writing the text myself

The legal-documents plan assumed I would author reviewed legal text describing what a site processes, across four jurisdictions and eight languages, and flagged that as the feature's main legal risk.

That framing was wrong in an important way. **I should not author claims about what a site processes on behalf of software I did not write.** WooCommerce's description of WooCommerce is both more accurate than anything I could write and, legally, WooCommerce's statement rather than mine. The generated document becomes the site's own installed software describing itself, plus the operator's own supplied facts, plus jurisdiction-fixed rights text — which is statute, not a claim about the site.

On faz-test, a single admin pageview picked up WordPress core, FAZ Cookie Manager and WooCommerce; with the rest of the site's plugin set active it also picks up LiteSpeed Cache, Akismet, Burst Statistics, WP Statistics and Beehive Analytics. None of that text is mine to write.

## The admin-only constraint, and the evidence for it

This is the decision most likely to be "simplified" away later by someone who does not know what it costs, so it is documented in three places: the class header, `docs/privacy-policy-collection-design.md`, and the commit message.

`wp_add_privacy_policy_content()` calls `_doing_it_wrong()` unless it runs in wp-admin on `admin_init` or later — the guard landed in 4.9.7. Producers therefore register on `admin_init`, and **their content simply does not exist in any other context**.

Forcing the issue is worse than useless. Firing `do_action( 'admin_init' )` from WP-CLI to warm the producers up emitted six `_doing_it_wrong` notices and then **fataled inside WooCommerce**: `Automattic\WooCommerce\Internal\Orders\OrderAttributionController::get_order_screen_id()` calls `wc_get_page_screen_id()`, which is undefined outside a real admin request.

So the hook is `current_screen`, not `admin_init`:

- `current_screen` is reached only through `set_current_screen()`, so it fires in genuine wp-admin page requests and nowhere else. `admin_init`, by contrast, also fires on `admin-ajax.php` and can be force-fired from the CLI — the exact path that fatals.
- It fires *after* `admin_init`, so every producer registered at default priority (FAZ's own included) has already contributed.

The context is therefore correct by construction. The explicit denials inside `maybe_collect()` — `is_admin()`, `did_action('admin_init')`, ajax, cron, `REST_REQUEST`, `WP_CLI`, `manage_options` — are belt and braces for anyone who calls the method directly.

Collection is further limited to screens whose id contains `faz-cookie-manager`. Running it on every wp-admin pageview sitewide would perform the read, and risk an option write, forever, in order to keep fresh a snapshot that only FAZ's document consumes. On FAZ screens it refreshes exactly when the operator is looking at the plugin — including, later, the privacy-policy editor screen itself, which lands on a `faz-cookie-manager-*` slug and so always opens against fresh data.

**Snapshot, not live**, follows from the same constraint — and is required independently anyway. Cache Compatibility Mode (1.21.0) demands render invariance, and a live collection at render time would make the output depend on request context, which is exactly the per-visitor variance that release eliminated.

## Editable-placeholder semantics

Mirrored from the Cookie Policy's `Section_Overrides` rather than reinvented:

- Collected text is the **placeholder**, not the saved value. An empty override box means "keep receiving updates", and the block silently adopts whatever the plugin now says.
- An operator who rewrites a block keeps their wording. The block stops tracking upstream, and an upstream change is **flagged, never silently applied** — the same judgement the Cookie Policy makes, because an explicit editorial decision is the most specific source there is.
- The anchor for that decision is the block's `source_hash` at the moment the override was saved.

What is genuinely new versus the Cookie Policy is that **the upstream source moves**. A plugin can change or withdraw its declaration, so each block carries `added` / `updated` / `removed` and a two-pass identity match:

1. **by text** — carries a block across a rename or a translated display name;
2. **by name** — carries it across a rewritten declaration.

A name that is ambiguous on both sides refuses to match. Two plugins can register under the same display name (which is why duplicate ids exist), and matching a rewrite against the wrong twin would silently swap two blocks' identities, and their overrides with them. Ambiguity makes the blocks churn instead, which is recoverable; swapped identities are not.

Three states are exposed by `describe()`, both **derived at read time and never stored** — a derived flag cannot desynchronise from the data it describes, a stored copy can (same reason `Section_Overrides` derives `active`):

| | untouched block | operator-edited block |
|---|---|---|
| plugin updated its text | adopts the new text | keeps operator wording, `stale` |
| plugin deactivated/removed | block drops out | kept, `orphaned` |
| new plugin activated | block appears | n/a |

I compute that diff myself instead of trusting core's. Core anchors its `added`/`updated`/`removed` to `wp_page_for_privacy_policy`, and on a site where that option is `0` there is no baseline and the flags are meaningless. Change detection has to work whether or not the operator ever configured a privacy page.

One trap in core's return value: entries carrying a `removed` timestamp are read back out of its post-meta cache and are **not** currently registered. Keeping them would resurrect every plugin the site ever deactivated, so they are dropped on the way in.

## The sanitisation boundary, and why it sits there

Each body goes through `wp_kses_post()`, then a character clip, then a trim — and `source_hash` is computed on that **final stored string**.

The order is load-bearing. Hashing the producer's raw text instead would make every block whose text kses touches hash differently from what is stored, producing a phantom `updated` — and a false `stale` flag on every operator-edited block — on every single collection, forever. The unit suite pins this with a deliberately observable kses stub (a naive `<script>` strip); weakening that stub to the identity function would make the suite pass without proving anything.

Operator overrides pass through the same `sanitize_html()`. There is one sanitisation path, not two.

Bounds (`MAX_BLOCKS` 100, `MAX_HTML` 60 000, `MAX_NAME` 200) exist so a pathological producer cannot grow the option without bound. The cap **refuses new blocks and never truncates the map** — truncation could evict a block the operator has edited, and losing an editorial decision is a worse failure than missing a section. The option is written with `autoload` false on every write path; the default for a new option is yes, so the third argument has to be passed every time, not only the first.

## What I tested

**Unit** — new `tests/unit/test-privacy-content-collector.php`, a self-contained CLI runner in the shape of the existing suites: **74 assertions, 74 passed, 0 failed**. Covers first collection, no-write idempotence, ghost/empty/nameless input rejection, kses-before-hash, untouched adoption, edited-block staleness, rename carry-forward, removal and revival, duplicate display names, `set_override` edges (unknown id, clearing, clipping, script stripping, deleting an orphan), every guard independently, corrupt-option recovery, and the `MAX_BLOCKS` cap.

`npm run test:unit`: **57 passed, 0 failed (of 57)** — 48 PHP + 9 JS, up from main's 47 + 9. No regression in any other suite.

**PHP syntax** — full sweep clean, including the four touched files.

**Live on faz-test**, per the test-before-claiming rule:

- one admin pageview as `admin` produced a `schema: 1` snapshot with three blocks — `wordpress`, `faz-cookie-manager`, `woocommerce` — each with non-empty `source_html` and a 64-character `source_hash`. (Only WooCommerce of the six third-party producers is currently active on that install; the others are installed but deactivated, and I did not activate them because a concurrent workflow is using the same site.) Worth noting: **WordPress core is itself a producer**, which was not in the expected list.
- two further FAZ admin pageviews, including a different FAZ screen, left `collected_at` byte-identical — the no-write path holds.
- `SELECT autoload FROM wp_options` → `off`, with the option at 10 162 bytes, which is exactly why it must not autoload.
- with the option deleted: a frontend request (200), a real `faz/v1` REST request and a WP-CLI run neither fataled nor recreated it. Collection provably never fires outside wp-admin.

**Plugin Check** (`--categories=plugin_repo`) caught a real regression I would otherwise have shipped: its direct-file-access scanner only reads the **first 50 lines** of a file when looking for the ABSPATH guard, and my original 85-line header docblock pushed the guard out of its view. I restructured the header — essentials up top, the long rationale in the class docblock below the guard and in the design doc — and the file now produces **zero findings of any kind**. Repo-wide errors went 398 → 397; every remaining one is pre-existing and lives in `tests/`, `scripts/`, `.specify/` or a dotfile, i.e. paths the wp.org ZIP excludes.

I did **not** run the Playwright E2E suite, or build the wp.org-shape ZIP for a clean-room Plugin Check run: a concurrent workflow is using the shared WordPress test site. The faz-test deployment and its option were reverted to their prior state after verification.

## Scope

Only `includes/class-cli.php` (four lines, in the method that already owns the producer-side registration — the existing `admin_init` producer hook is correct as it stands and I did not move it) and `uninstall.php` (one option name). **No file under `admin/modules/cookie-policy-generator/` is touched**, and no call to `wp_add_privacy_policy_content()` is added or moved.

## Follow-ups

- **`Document_Registry` integration**, blocked on the concurrent branch: the privacy-policy `Document_Config` gains `'collector' => Content_Collector::collect` and `'snapshot_option' => 'faz_privacy_content_snapshot'`. No migration needed — `effective_blocks()` already returns render-ready `id => [plugin_name, html]` pairs.
- **Renderer**: the public document built from `effective_blocks()`, never a live collection, through the same role/aria-level-augmented `wp_kses` allowlist the Cookie Policy renderer uses.
- **Admin screen** (`faz-cookie-manager-privacy-policy`): per-block editor with `source_html` as a greyed placeholder, an override textarea, and the three-state flags. Its slug triggers collection on open by construction.
- **Offer — never silently set** — to point `wp_page_for_privacy_policy` at the generated page, so `get_privacy_policy_url()` and `Renderer::privacy_policy_url()` resolve sitewide.
- Jurisdiction-fixed rights scaffolds (GDPR Arts. 15-22, CCPA, LGPD Art. 18, POPIA) and the structural frame around collected blocks, with `missing_required_settings()`-style refusal gating and the never-seed-from-`admin_email` rule.
- `{{COMPANY_NAME}}` resolution inside operator overrides, through the same substitution stage as the Cookie Policy so there is no second escaping path.
- An admin-context refresh button. REST is not an option here unless proxied through a real admin request.
- i18n resync once the UI strings land — one pass after all concurrent branches merge.

## Summary by CodeRabbit

* **Nuove funzionalità**
  * Raccolta automatica dei contenuti privacy dichiarati dai plugin nelle schermate amministrative autorizzate.
  * Salvataggio di uno snapshot aggiornabile per generare documenti privacy coerenti e verificabili.
  * Gestione di contenuti modificati, rimossi o non più disponibili, con possibilità di applicare override editoriali.
  * Validazione, sanificazione e limiti dimensionali per migliorare affidabilità e sicurezza dei contenuti.

* **Documentazione**
  * Aggiunta la documentazione sul funzionamento della raccolta e sulla gestione dei contenuti.

* **Manutenzione**
  * Lo snapshot viene rimosso automaticamente durante la disinstallazione.

### DIFF
--- a/admin/modules/privacy-policy-generator/includes/class-content-collector.php
+++ b/admin/modules/privacy-policy-generator/includes/class-content-collector.php
@@ -1,0 +1,712 @@
+<?php
+/**
+ * Collect what the installed plugins declare about their own data processing.
+ *
+ * Since 4.9.6 a plugin declares its own processing by calling
+ * `wp_add_privacy_policy_content( $plugin_name, $policy_text )`, and
+ * `WP_Privacy_Policy_Content::get_suggested_policy_text()` hands back every
+ * registered contribution. FAZ is already a producer of one such block (see
+ * `CLI::register_privacy_hooks()`); this class makes it a consumer too.
+ *
+ * HARD CONSTRAINT — collection is admin-only, and this is not negotiable.
+ * `wp_add_privacy_policy_content()` calls `_doing_it_wrong()` unless it runs
+ * inside wp-admin on `admin_init` or later (guard added in 4.9.7), so
+ * producers register there and their content simply does not exist in any
+ * other context. Forcing the issue is worse than useless: firing
+ * `do_action( 'admin_init' )` from WP-CLI to "warm up" the producers emitted
+ * six `_doing_it_wrong` notices and then fataled inside WooCommerce, whose
+ * `OrderAttributionController::get_order_screen_id()` calls
+ * `wc_get_page_screen_id()` — undefined outside a real admin request. Do not
+ * move this onto a REST route or a cron event later; it will fatal on sites
+ * running WooCommerce.
+ *
+ * SNAPSHOT, NOT LIVE. Collection therefore cannot happen at render time: the
+ * blocks are persisted into the `faz_privacy_content_snapshot` option and the
+ * public document renders from that. Required independently anyway — Cache
+ * Compatibility Mode (1.21.0) demands render invariance, and a live collection
+ * would make the output depend on request context.
+ *
+ * Full rationale: docs/privacy-policy-collection-design.md.
+ *
+ * @package FazCookie\Admin\Modules\Privacy_Policy_Generator\Includes
+ * @since   1.26.0
+ */
+
+namespace FazCookie\Admin\Modules\Privacy_Policy_Generator\Includes;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Snapshot the privacy-policy content other plugins register about themselves.
+ *
+ * The point of aggregating is editorial, not technical: we must not author
+ * claims about what a site processes on behalf of software we did not write.
+ * WooCommerce's description of WooCommerce is both more accurate than anything
+ * we could write and, legally, WooCommerce's statement rather than ours.
+ *
+ * EDITABLE PLACEHOLDERS, mirroring `Section_Overrides`. The collected text is
+ * a placeholder, not a saved value: an empty override means "keep tracking
+ * upstream" and the block silently adopts whatever the plugin now says. Once
+ * the operator writes their own wording, that wording is the document and an
+ * upstream change is flagged, never silently applied — the same judgement the
+ * Cookie Policy makes, because an explicit editorial decision is the most
+ * specific source there is. The anchor for that decision is the block's
+ * `source_hash` at the moment the override was saved; the block's own identity
+ * anchor is the plugin, matched first by text (so a rename or a translated
+ * display name carries the block forward) and then by name (so a rewritten
+ * declaration updates the block it belongs to).
+ *
+ * FAZ keeps its own snapshot and computes its own diff on purpose. Core's
+ * `added`/`updated`/`removed` flags are computed against the
+ * `_wp_suggested_privacy_policy_content` post meta of the page named by
+ * `wp_page_for_privacy_policy`; on a site where that option is 0 there is no
+ * baseline and those flags are meaningless. Change detection has to work
+ * whether or not the operator ever configured core's privacy page.
+ *
+ * KNOWN, ACCEPTED SIDE EFFECT. `get_suggested_policy_text()` refreshes core's
+ * `_wp_suggested_privacy_policy_content` meta cache when a privacy page IS
+ * configured, which can pre-empt core's own "suggested text has changed" admin
+ * notice. It is core's public, sanctioned accessor — the same refresh happens
+ * whenever the operator opens core's Privacy Policy Guide — the effect is
+ * bounded to installs that both configured a privacy page and visit FAZ
+ * screens, and FAZ's own three-state flagging supersedes that notice for
+ * FAZ-managed documents. Reading the private `$policy_content` property by
+ * Reflection to avoid it was considered and rejected: worse for wp.org review
+ * and brittle against core refactors.
+ *
+ * KNOWN BLIND SPOT, documented rather than fixed. If a site's admin locale
+ * changes, every translated `policy_text` changes with it, so every block
+ * legitimately reports an update. Worse, a plugin that translates BOTH its
+ * name and its text matches neither identity pass and cycles removed + added,
+ * orphaning any override on it. Core has exactly the same limitation, and
+ * plugin display names are in practice untranslated brand names.
+ */
+class Content_Collector {
+
+	/**
+	 * Option holding the snapshot. Written with autoload false: it is
+	 * admin/render-source data, never needed on an ordinary frontend request,
+	 * and can run to tens of kilobytes on a plugin-heavy site.
+	 */
+	const OPTION = 'faz_privacy_content_snapshot';
+
+	/** Stored shape version. Bump and migrate in place when the shape changes. */
+	const SCHEMA = 1;
+
+	/**
+	 * Hard cap on stored blocks. No real install comes close; the cap exists
+	 * so a malformed or pathological producer cannot grow the option without
+	 * bound. Same rationale as `Section_Overrides::MAX_SECTIONS`.
+	 */
+	const MAX_BLOCKS = 100;
+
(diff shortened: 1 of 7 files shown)


fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
Thanks for looking. I checked this one empirically before changing anything, and it does not hold — no `require_once` is needed.

The plugin ships a convention-based autoloader (`class-autoloader.php`), registered in the bootstrap at `faz-cookie-manager.php:447-448`, before anything else runs. `Autoloader::load_class()` maps a `FazCookie\*` class name to a path mechanically: lowercase each segment, `_` → `-`, and prefix the last one with `class-`. So

`FazCookie\Admin\Modules\Privacy_Policy_Generator\Includes\Content_Collector`

resolves to

`admin/modules/privacy-policy-generator/includes/class-content-collector.php`

which is exactly where the file is. Verified by loading the autoloader in isolation and asserting resolution:

```
Content_Collector risolta: SI
Renderer (controllo)      : SI
```

The reason the search came back empty is that this is not Composer PSR-4 — there is no `autoload` block in `composer.json` to find, and the mapping lives in that one custom `spl_autoload_register` callback.

The wider point is fair though: the module is new and nothing else in it had proven the convention yet. That is now covered by the unit suite, which instantiates the class through the same path.

PR #204, opened 2026-07-31
feat: opt-in footer legal links, consent-invariant and cache-safe

fabiodalez-dev:
## What this adds

A default-OFF setting that lets the admin pick which published pages get printed as a small legal-links `<nav>` in the theme footer (`wp_footer`, priority 20, so it lands after the consent banner). Typical use: "Cookie Policy · Privacy Policy · Imprint" without editing a theme template or dropping a shortcode into a widget area.

```html
<nav class="faz-legal-links" aria-label="Legal information">
  <ul class="faz-legal-links-list">
    <li class="faz-legal-links-item"><a href="…">Cookie Policy</a></li>
  </ul>
</nav>
```

## Design decisions and why

**It lives in `faz_settings`, not its own option.** My plan (§4.1) parked this in a standalone `faz_legal_links` option tied to the future Legal Documents registry. I deliberately deviated so this can ship on its own: the existing REST save path (`POST faz/v1/settings` → `Store::update` → `Settings::sanitize`) already handles persistence, so there is no new endpoint, no new capability check and no new nonce surface. The whole server side is one `get_defaults()` entry, one `get_excludes()` entry and one `sanitize_option()` case. No Legal Documents admin page is built here.

**The key is `link_items`, not a generic `items`/`pages`.** `get_excludes()` and `sanitize_option()` match by *bare key name across the entire settings tree*, so a generic name would hijack any future group that happened to reuse it.

**The `get_excludes()` entry is load-bearing.** The recursive sanitiser walks arrays-of-structs against their default, and the default here is an empty array — without the exclusion, every stored row is wiped on the next save of *any* setting. Same trap the existing `variants` entry documents. The round-trip assertion in the new unit suite exists purely to catch that: I verified it by temporarily removing the `link_items` line from `get_excludes()` and confirming the test goes red, then restoring it.

**Invariance is the whole point.** `build_html()` reads nothing request-dependent — no `$_COOKIE`, no consent store, no geo lookup, no `is_user_logged_in()` branch. The output is a pure function of the option plus each page's publish state. Cache Compatibility Mode promises one cached variant per URL, and 1.21.0 only shipped once every render path had been made invariant; I did not want this to be the first exception. The rule is stated in the method docblock so a future edit has to argue with it. The E2E spec asserts the nav's `outerHTML` is byte-identical before and after the visitor accepts.

**Unpublished pages are skipped at render time**, not pruned on save, so unpublishing a page removes its footer link immediately with no admin round trip. A blank label falls back to the live page title for the same reason: renaming the page keeps the link in sync. I used `get_the_title()` rather than `$post->post_title` so translation plugins can localise the label — that varies per URL (WPML/Polylang serve each language on its own URL), never per visitor, so the one-variant-per-URL guarantee still holds.

**`(int)` instead of `absint()` on `page_id`**, in both the sanitiser and the renderer. `absint(-5)` is `5`, so a malformed ID would silently resolve to a *different, real* page. Malformed input has to drop out, not get rounded into something valid. This is the one place I diverged from my own spec snippet, and the unit test pins the behaviour.

**Bounds:** 20 rows max, labels clipped to 120 chars (`mb_substr` when mbstring is present, `substr` otherwise). A footer nav is a handful of legal pages, and the cap keeps a crafted settings PUT from bloating every cached page on the site. The renderer carries the same cap defensively, for rows that predate it or arrived via a direct DB edit.

**Zero surviving links renders nothing at all** — an empty `<nav>` is a landmark with no content for assistive technology.

**CSS is enqueued from `wp_enqueue_scripts`, never from `render()`.** By `wp_footer` time `wp_head` is long gone and a late enqueue would be dropped or printed illegally in `<body>`. It is also gated on the feature actually being on, so an untouched install ships zero extra bytes. The stylesheet is ~15 lines, colourless, no `!important` — a theme that wants to restyle the nav should win without a specificity fight.

**Admin UI:** the page rows are server-rendered and carry **no `data-path`** — a stray one would coerce each row into a boolean under `legal_links` and corrupt the payload. `settings.js` hydrates and serialises them through its own helpers. Because the rows are server-rendered there is no async-readiness guard to write, unlike the A/B-test variant list. The one guard I did add covers a site with no published pages at all, where the view renders no rows and a naive serialize would report "nothing selected" and wipe a list the admin configured while pages still existed. `enabled` (via `data-path`) and `link_items` (via the custom serializer) are always sent together, since the save path merges group objects with `Object.assign`.

## Backward compatibility

- Default is `enabled => false` with an empty `link_items`, so an install upgrading to this version renders no nav and enqueues no CSS. Nothing changes until an admin opts in.
- No DB migration, no new option, no new table, no new REST route.
- No outbound HTTP, no CDN assets — the CSS ships in the ZIP.
- No frontend JS touched, so `npm run build:min` was not needed.
- One existing test file changed: `test-flyingpress-frontend-bootstrap-php.php` constructs the real `Frontend` without an autoloader, so every collaborator has to be declared there; it gained a one-line `Legal_Links` stub. Nothing about what that suite asserts changed.

## What I tested

- `npm run test:unit` — **57 passed, 0 failed (of 57)**. That is the previous 56 plus the new `test-legal-links-php.php`; nothing turned red, and `test-cookie-policy-golden-render.php` (the backward-compatibility gate) is untouched and green.
- New `tests/unit/test-legal-links-php.php` — **20 assertions, all passing**: default OFF, empty selection, draft-only selection, title fallback, custom label, label markup escaped to entities, permalink through `esc_url`, title through `esc_html`, draft/trashed/deleted/untitled pages skipped, stored order preserved, byte-identical output across anonymous/accepted/rejected consent cookies, plus the sanitiser (non-array collapse, dedupe, zero and negative IDs dropped, unknown keys dropped, 120-char clip, 20-row cap) and the `sanitize()` round trip.
- PHP syntax sweep over the whole repo (excluding `vendor/`, `node_modules/`, `index.php`) — clean.
- `npx tsc --noEmit --strict` on the new spec — clean. `npx eslint` on `settings.js` — clean.
- Verified the autoloader maps `FazCookie\Frontend\Modules\Legal_Links\Legal_Links` to `frontend/modules/legal-links/class-legal-links.php`.
- Verified the get_excludes regression guard actually bites, as described above.

## What I did not do

- **I did not run the Playwright E2E suite**, per the house rule that it drives a single shared WordPress test site. For the same reason I did not rsync this branch over the shared deploy path — a `--delete` sync there would have clobbered whatever else is currently deployed. `tests/e2e/specs/legal-footer-links.spec.ts` and the existing `cache-compatibility-mode.spec.ts` regression check still need to be run in the separate E2E pass.
- No i18n resync (`make-pot`/`msgmerge`). New translatable strings are present; the standing rule is one resync per release.

## Summary by CodeRabbit

- **Nuove funzionalità**
  - Aggiunta la sezione “Footer legal links” nelle impostazioni.
  - È possibile selezionare fino a 20 pagine pubblicate e personalizzare le relative etichette.
  - I collegamenti legali selezionati vengono mostrati nel footer del sito.

- **Correzioni**
  - Le pagine non pubblicate o non valide vengono escluse automaticamente.
  - Le impostazioni esistenti vengono preservate quando non sono presenti selezioni.

- **Test**
  - Aggiunti test end-to-end per visualizzazione, disattivazione e personalizzazione dei collegamenti.

### DIFF
--- a/admin/assets/js/pages/settings.js
+++ b/admin/assets/js/pages/settings.js
@@ -229,37 +229,19 @@
 	 * [{ page_id, label }]. DOM order is the rendered order, which is what the
 	 * admin sees on screen.
 	 *
-	 * Entries whose page has no row on this screen are carried over untouched.
-	 * The view renders only published pages and only the first 200 of them, so a
-	 * stored page can legitimately have no checkbox — and an absent checkbox is
-	 * not the admin unticking it, it is the admin never having been shown it.
-	 * Inferring "unchecked" from "not rendered" would silently delete links on
-	 * save. Carrying them over also matches how the renderer treats publish
-	 * state: the stored list is the admin's intent, publish state is applied at
-	 * render time, so unpublishing hides a link and republishing brings it back.
-	 *
-	 * @param {Array} stored The currently stored link_items, used as the source
-	 *                       for entries that have no row on screen.
+	 * The PHP view guarantees a row for every stored selection, including a
+	 * deleted/unpublished page or one beyond the published-page query cap. An
+	 * unchecked row can therefore always mean an explicit removal.
 	 */
-	function serializeLegalLinks(stored) {
+	function serializeLegalLinks() {
 		var items = [];
 		var container = document.getElementById('faz-legal-links-pages');
 		if (!container) return items;
-		var rendered = {};
-		container.querySelectorAll('input.faz-legal-link-page').forEach(function (cb) {
-			rendered[String(cb.value || '')] = true;
-		});
 		container.querySelectorAll('input.faz-legal-link-page:checked').forEach(function (cb) {
 			var pageId = parseInt(cb.value, 10);
 			if (!pageId) return;
 			var labelInput = container.querySelector('input.faz-legal-link-label[data-page-id="' + cb.value + '"]');
 			items.push({ page_id: pageId, label: ((labelInput && labelInput.value) || '').trim() });
-		});
-		(Array.isArray(stored) ? stored : []).forEach(function (item) {
-			if (!item) return;
-			var pageId = parseInt(item.page_id, 10);
-			if (!pageId || rendered[String(pageId)]) return;
-			items.push({ page_id: pageId, label: typeof item.label === 'string' ? item.label : '' });
 		});
 		return items;
 	}
@@ -369,10 +351,7 @@
 				? current.legal_links.link_items
 				: [];
 			if (document.getElementById('faz-legal-links-pages')) {
-				// The stored list is passed in so entries whose page is not among the
-				// rendered rows (unpublished, or beyond the view's 200-page cap)
-				// survive the save instead of being read as "unticked".
-				formData.legal_links.link_items = serializeLegalLinks(storedLegalLinks);
+				formData.legal_links.link_items = serializeLegalLinks();
 			} else {
 				// The site has no published pages, so the view rendered no rows at
 				// all — serializing would report "nothing selected" and wipe a list
--- a/admin/assets/js/pages/settings.js
+++ b/admin/assets/js/pages/settings.js
@@ -228,16 +228,38 @@
 	 * Collect the checked footer legal links, in DOM order, as
 	 * [{ page_id, label }]. DOM order is the rendered order, which is what the
 	 * admin sees on screen.
+	 *
+	 * Entries whose page has no row on this screen are carried over untouched.
+	 * The view renders only published pages and only the first 200 of them, so a
+	 * stored page can legitimately have no checkbox — and an absent checkbox is
+	 * not the admin unticking it, it is the admin never having been shown it.
+	 * Inferring "unchecked" from "not rendered" would silently delete links on
+	 * save. Carrying them over also matches how the renderer treats publish
+	 * state: the stored list is the admin's intent, publish state is applied at
+	 * render time, so unpublishing hides a link and republishing brings it back.
+	 *
+	 * @param {Array} stored The currently stored link_items, used as the source
+	 *                       for entries that have no row on screen.
 	 */
-	function serializeLegalLinks() {
+	function serializeLegalLinks(stored) {
 		var items = [];
 		var container = document.getElementById('faz-legal-links-pages');
 		if (!container) return items;
+		var rendered = {};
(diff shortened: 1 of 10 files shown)


fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
Thanks — both findings were fair, and both are now fixed in 451d8f8. Notes on how I applied them:

**`get_pages()` unbounded (`admin/views/settings.php`)** — agreed, and capped at 200.

I did not apply the suggested diff as-is, because the cap on its own would have introduced a silent data-loss bug. `serializeLegalLinks()` in `settings.js` reads the rendered checkboxes and treats anything unticked as removed. Once the list is capped, a stored page falling outside the rendered rows would be read as "unticked" and dropped from the saved list — on any settings save, not just an edit to this card. An absent checkbox is not the admin unticking a link, it is the admin never having been shown it, so the serializer now carries over stored entries that have no row on screen. That also fixes the pre-existing version of the same problem when a selected page is unpublished, and it lines up with how the renderer already behaves: the stored list is the admin's intent, publish state is applied at render time, so unpublishing hides a link and republishing brings it back.

I left `sort_column` out — `post_title` is already the `get_pages()` default, so passing it explicitly would not change the ordering.

**`maybe_enqueue_styles()` not checking that anything renders** — agreed, and fixed the way you suggested, with the memoised result shared between the enqueue pass and `render()`.

Worth noting the cost argument runs in the right direction here: building at `wp_enqueue_scripts` and reusing the memo at `wp_footer` means the page lookups happen once per request instead of twice, so the stricter check is cheaper than the one it replaced. A site that never enabled the feature still short-circuits on the `enabled` flag before touching a single page.

I added six assertions for the gate: off, empty list, every page unpublished or deleted, a page that really renders, admin requests, and one pinning that `render()` after the enqueue pass still prints the freshly built markup, so the memo cannot drift from a fresh build. I checked the unpublished-pages assertion actually bites by restoring the old option-based check, which turned it red.

Unit suites 57 passed / 0 failed, PHP syntax sweep clean, eslint clean.

PR #207, opened 2026-08-01
Update Czech translation for v1.25

vvvamik:
Updated translation based on v1.25
Updated both .po and .mo files.
Testing: Tested locally on localhost (v1.25) – all Czech strings render correctly without issues.


fabiodalez-dev:
Thank you — this is a big improvement for Czech: 1170 translated strings with 63 fuzzy and 213 missing, up to 1421 translated with none fuzzy and only 6 left.

I checked the files rather than assuming, and two things are worth saying because most translation PRs get them wrong:

- The `.mo` is byte-identical to a fresh `msgfmt` of your `.po`. That is the file WordPress actually loads at runtime, and a stale or hand-edited one silently ships the old strings.
- `msgfmt --check` is clean: no broken placeholders, no bad escapes.

One thing I need to do before merging, and it is not a problem with your work. You translated against the 1.25 POT, and two features landed after it: the footer legal links and the policy version ledger. Merging your file as it stands would drop the 13 msgids they introduced, so I will run `msgmerge` against the current POT and recompile the `.mo`.

I verified that keeps 1419 of your 1421 translations and restores all 13 missing strings. The two that drop out are msgids no longer in the POT — strings removed from the code, not work lost.

The 27 that remain untranslated afterwards are those 13 new ones plus the ones that were already missing; no action needed from you.

PR #208, opened 2026-08-02
Stop granular consent mode from deleting cookies of accepted categories

fabiodalez-dev:
## The bug

With per-service or per-cookie consent enabled, a cookie belonging to an **accepted** category was deleted client-side unless it carried a redundant `svc.<id>:yes` token. A visitor who consented to analytics had their analytics cookies shredded anyway.

Present in released versions. Found while investigating a pre-existing E2E failure that looked like a test artefact.

## Mechanism

In granular mode `_fazRemoveAllDeadCookies()` deliberately sweeps **every** category, including accepted ones, so an explicit `svc.*:no` or `ck.*:no` inside an accepted category is enforced immediately instead of waiting for the next request's server-side shred.

`_fazRemoveDeadCookies()` checked those explicit decisions correctly, but when `_fazGetServiceCookieDecision()` returned `""` — no override at all — it went straight to deleting any cookie it found rather than consulting the category. A service that simply follows its accepted category has no reason to carry a redundant allow token, so **absence of a token was read as denial**.

That inverted the documented precedence:

1. explicit per-cookie (`ck.*`)
2. explicit per-service (`svc.*`)
3. category consent

The fix restores step 3: no explicit decision plus a category that is not blocked means keep. One line.

Necessary cookies are preserved by the same change rather than by separate handling — `_fazIsCategoryToBeBlocked()` already treats an unset store value as "keep" for categories flagged `isNecessary`, and it is now actually consulted, so the first cleanup pass before the consent store is populated no longer wipes them.

## Why it stayed invisible

Stripe was the only provider that appeared to behave, and only because its administrator-enabled gateway cookie patterns are whitelisted. That exemption masked the same missing fallback rather than making the behaviour correct.

## Scope, stated plainly

Reachable only when granular consent is enabled — `per_service_consent` and `per_cookie_consent` both default to **off**.

It **over-deletes rather than under-deletes**, so no visitor data was ever exposed and this is not a compliance failure. It breaks analytics for visitors who consented, which is a different kind of serious.

## Tests

`tests/unit/js/granular-cookie-shredder.test.mjs` pins both directions, so the fix cannot be reintroduced *or* over-applied:

- accepted category keeps its cookie without a redundant allow ← the regression
- denied category still loses its cookie without an explicit deny ← the opposite failure
- explicit `svc.*` / `ck.*` decisions still override the category in both directions
- Necessary survives the cleanup pass that runs before the consent store is populated

**Verified non-vacuous:** removing the one-line fix turns 2 of its 6 checks red; restoring it returns 6/6.

## Verification

- `blocking-compliance.spec.ts` — **7/7**, including all 16 consent-matrix combinations. Previously the first combination that awaited any cookie timed out after 30s.
- `npm run test:unit` — 62/62, with the new suite actually picked up by the runner.
- `script.min.js` regenerated and confirmed byte-identical to a fresh `npm run build:min`.

## Release note

1.26.0 is prepared on `main` but not tagged, and wp.org still serves 1.25.0. This fix can still go into it, which is my recommendation: the defect is already in released versions and the fix is contained and verified.

## On the commit history

The first commit on this branch records a diagnosis that turned out to be **wrong** — it claimed the provider scripts were absent from the page rather than blocked. They were present and executing. The second commit corrects the document with the verified mechanism. Left both rather than rewriting, since the wrong turn is part of how the real cause was found.

## Summary by CodeRabbit

- **Correzioni**
  - Corretta la rimozione automatica dei cookie per rispettare le categorie consentite quando non sono presenti decisioni più specifiche.
  - Preservate le autorizzazioni esplicite a livello di servizio o di singolo cookie.
  - Gestita correttamente l’eccezione relativa ai cookie Stripe.

- **Documentazione**
  - Aggiunta documentazione sul problema, sulla soluzione prevista e sui test di regressione.


fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev (commented):
## Release review completed

### Findings and corrections
- Reviewed granular cookie shredding precedence from category decision through service and per-cookie overrides.
- Confirmed that the production fix preserves cookies from accepted categories and that explicit narrower consent decisions retain precedence.
- Found that the new suite did not explicitly lock the most specific allow case: a per-cookie allow must beat both service-level and category-level denial.
- Added that regression in commit 566e8e9, completing the precedence matrix without changing production behavior.

### Validation
- Granular shredder suite: 7/7.
- Full unit suite: 62/62.
- Blocking end-to-end suite: 7/7, including all 16 category combinations.
- Minified frontend build: byte-identical.
- PHPStan: clean.

frontend/js/script.js
```diff
@@ -2073,9 +2073,10 @@ function _fazSetFooterShadow($doc) {
 /**
  * Remove all the rejected cookies.
  *
- * @param {object} cookies Cookies list.
+ * @param {object} category Category with slug and cookies list.
  */
-function _fazRemoveDeadCookies({ cookies }) {
+function _fazRemoveDeadCookies(category) {
```
fabiodalez-dev, line 2078: **question:** The parameter contract just widened in a way that fails silently if a caller wasn't updated.

Before, this accepted **anything with a `cookies` array** — the destructuring `{ cookies }` didn't care where the object came from. Now it needs a real category with a usable `.slug`, because that slug feeds `_fazIsCategoryToBeBlocked()` two lines down.

> If some call site still passes a synthesized `{ cookies: [...] }` (a merged list, a service-shaped object, a test double), then `category.slug` is `undefined`, `_fazIsCategoryToBeBlocked(undefined)` almost certainly returns falsy, and **every cookie in that call is kept unconditionally** — including ones the visitor explicitly refused at category level.

That's the mirror image of the bug you're fixing, and it's the direction that *is* a compliance problem. It also fails quietly: no throw, no console warning, just cookies that don't get removed.

Have you grepped all call sites? The diff only shows the two comment updates in `_fazAcceptCookies()` and `_fazRemoveAllDeadCookies()`, so I can't tell from here whether those are the only two callers.

If they are, this is fine as-is. If you want a cheap belt-and-braces guard, a slug-less category is a programming error rather than a runtime state, so failing loudly beats defaulting either way.

frontend/js/script.js
```diff
@@ -2086,6 +2087,11 @@ function _fazRemoveDeadCookies({ cookies }) {
         // whitelist only overrides the category fallback when no explicit
         // decision exists.
         if (serviceDecision === "yes") continue;
+        // No granular override means the category remains authoritative. In
+        // granular mode callers intentionally sweep every category so an
+        // explicit svc/ck denial inside an accepted category is enforced; that
+        // must not turn a missing override into an implicit denial.
```
fabiodalez-dev, line 2093: **suggestion:** The fix itself reads correctly, and I like that it's placed **before** the whitelist check rather than after — for an accepted category both branches keep the cookie, and for a blocked category the whitelist still gets its say, so the ordering preserves the Stripe exemption exactly where it's still meaningful.

Minor: `_fazIsCategoryToBeBlocked(category.slug)` is category-scoped but re-evaluated once per cookie. For a category with many cookies that's a pointless repeat of the same store lookup. Hoisting it above the loop would also make the invariant more obvious to the next reader — *the category decision is fixed for the whole sweep, only the per-cookie override varies*:

```js
const cookies = category.cookies || [];
const categoryBlocked = _fazIsCategoryToBeBlocked(category.slug);
```

then `if (serviceDecision !== "no" && !categoryBlocked) continue;`

Not worth blocking on — the current form is correct, and I'd rather ship the one-line fix than grow it.

tests/unit/js/granular-cookie-shredder.test.mjs
```diff
@@ -0,0 +1,248 @@
...
+import { JSDOM } from 'jsdom';
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import { dirname, resolve } from 'node:path';
+
+const HERE = dirname(fileURLToPath(import.meta.url));
+const SCRIPT_PATH = resolve(HERE, '../../../frontend/js/script.js');
+
+let passed = 0;
+let failed = 0;
```
fabiodalez-dev, line 19: **bug:** The harness counts failures but I can't see from the truncated patch whether anything ever *acts* on `failed`.

If the file ends without something like:

```js
if (failed > 0) process.exitCode = 1;
```

then every assertion can go red and `npm run test:unit` still exits `0`. The suite would print `FAIL` lines into a green CI log — the worst possible outcome for a regression guard, because it looks like it's protecting the fix while protecting nothing.

You verified non-vacuousness by hand (2 of 6 checks flip when the fix is removed), which is exactly the right check to have run — but that verification only holds if the runner propagates the result. Worth confirming the final lines do it, and that the `62/62` you report would actually have been `60/62` rather than a silent pass.

tests/unit/js/granular-cookie-shredder.test.mjs
```diff
@@ -0,0 +1,248 @@
...
+    ...overrides,
+  };
+
+  const realAdd = window.document.addEventListener.bind(window.document);
+  window.document.addEventListener = (type, ...rest) => {
+    if (type === 'DOMContentLoaded') return undefined;
+    return realAdd(type, ...rest);
+  };
+  window.eval(readFileSync(SCRIPT_PATH, 'utf8'));
+  window.document.addEventListener = realAdd;
```
fabiodalez-dev, line 62: **question:** Nice touch suppressing only `DOMContentLoaded` and restoring the real `addEventListener` immediately after `window.eval()` — that keeps the bootstrap from firing without leaving the document permanently monkey-patched for the assertions that follow.

One question on fidelity: `_fazConfig` here is hand-authored, so the suite pins the shredder against a *shape you wrote*, not against the shape the plugin actually emits. If PHP ever renames `_cookieCategoryMap` or starts omitting `isNecessary` on categories, this suite keeps passing while production breaks.

Not something to fix in this PR — but is there an existing fixture or a serialized real config elsewhere in `tests/` that these unit tests could share, so the two definitions can't drift apart?

fabiodalez-dev (commented):
### Verdict
**looks good to me**

### Primary Goal
Restore the documented consent precedence in the client-side cookie shredder so that, in granular mode, a cookie whose category was accepted is no longer deleted just because it carries no redundant `svc.*:yes` token.

### Overview
### What this PR does

One behavioural line in `frontend/js/script.js`, plus the supporting signature change, a regression suite, and the release paperwork.

- `_fazRemoveDeadCookies()` now takes the **whole category object** instead of destructuring `{ cookies }`, so it can consult `category.slug`.
- New guard: `if (serviceDecision !== "no" && !_fazIsCategoryToBeBlocked(category.slug)) continue;` — absence of an explicit `svc.*` / `ck.*` decision falls back to category consent instead of being read as denial.
- Two stale comments in `_fazAcceptCookies()` and `_fazRemoveAllDeadCookies()` are corrected: they claimed the helper kept *only* cookies whose effective decision is `yes`, which was exactly the wrong invariant being relied on.
- `tests/unit/js/granular-cookie-shredder.test.mjs` loads the real `script.js` under jsdom with the `DOMContentLoaded` bootstrap suppressed and pins both directions (accepted keeps, denied still deletes, explicit overrides still win, necessary survives the pre-store pass).
- `CHANGELOG.md`, `readme.txt`, and a new `docs/` note record the defect and the mechanism.

### Why the fix is the right shape

The interesting part is that necessary cookies are preserved *as a consequence* rather than by a special case: `_fazIsCategoryToBeBlocked()` already treats an unset store value as "keep" for `isNecessary` categories, and it is now actually reached. That's one behaviour with one owner instead of two code paths that can drift.

The new guard is also placed **before** the whitelist check, which is correct: for an accepted category both branches keep the cookie, and for a blocked category the whitelist still gets its say.

### Concerns

Nothing blocking. Two things I'd like confirmed before merge — see inline:

1. The parameter contract changed from "anything with `.cookies`" to "a real category with `.slug`". Any caller still passing a synthesized `{ cookies: [...] }` would now silently hit the *opposite* failure mode (under-deletion), which unlike the bug being fixed **is** a compliance concern.
2. The hand-rolled test harness needs to set a non-zero exit code on failure, otherwise the suite is decorative in CI.

### Scope and risk

Tight and proportionate. The claim that this over-deletes rather than under-deletes holds for the change as written — the new guard only ever adds a `continue`, it never removes one — so the blast radius is "cookies that used to be wrongly deleted now survive", which is the intent.

### Scope Assessment
In scope. Every file maps to the stated goal: one behavioural line in `frontend/js/script.js`, the regenerated `script.min.js`, a regression suite, and changelog/readme/doc entries. The two comment rewrites in `_fazAcceptCookies()` and `_fazRemoveAllDeadCookies()` are not drift — they documented the exact wrong invariant that made the bug look intentional, so leaving them would be worse than touching them. `docs/blocking-compliance-analytics-timeout.md` is a new file that also narrates the investigation (including the initial wrong diagnosis); that's borderline for a `docs/` directory versus a PR description, but it's additive and harmless.

### Risk Assessment
The change can only *keep* cookies that were previously deleted, never the reverse, so it cannot introduce an under-deletion regression **on the paths shown**. The residual risk is entirely in the signature change: `_fazRemoveDeadCookies` went from accepting any `{ cookies }`-shaped object to requiring a category with a usable `.slug`. A caller passing a slug-less object would get `_fazIsCategoryToBeBlocked(undefined)` — almost certainly falsy — and therefore keep every cookie unconditionally, silently, including after an explicit category refusal. That failure mode is the compliance-relevant one, and it would be invisible: no throw, no log. Worth one grep across call sites. Secondary risk: if the new jsdom harness doesn't fail the process on assertion failure, the regression guard silently stops guarding.

### Reuse Notes
- `_fazIsCategoryToBeBlocked()` is correctly reused as the single source of truth for category state, including the `isNecessary` + unset-store case — no parallel "is this necessary" branch was added, which is the right call.
- The whitelist path (`_fazIsCookieWhitelisted`) is left intact and correctly ordered after the new category fallback, so the Stripe payment-gateway exemption keeps working inside genuinely blocked categories.
- The test loads the shipped `frontend/js/script.js` directly rather than re-implementing the precedence logic, so it can't drift from production behaviour — good pattern to keep for future consent-logic tests.

### Action Items
- [ ] Grep every call site of `_fazRemoveDeadCookies` and confirm all of them pass a real category object with `slug`, not a synthesized `{ cookies: [...] }`.
- [ ] Confirm the new jsdom suite sets `process.exitCode = 1` (or throws) when `failed > 0`, so `npm run test:unit` actually goes red.
- [ ] Optional: hoist `_fazIsCategoryToBeBlocked(category.slug)` out of the per-cookie loop — it's category-scoped and re-evaluated per cookie.

PR #209, opened 2026-08-02
Resync the Czech catalogue to the current POT

fabiodalez-dev:
Follow-up to #207, which contributed a Czech translation made against the 1.25 POT.

Two features landed after 1.25 — the footer legal links and the policy version ledger — so the contributed file did not carry the 13 msgids they introduced. Merging it verbatim removed those strings from the catalogue; `msgmerge` against the current POT restores them.

- **1419 of the 1421 contributed translations are kept.** The two that drop out are msgids no longer in the POT — strings removed from the code, not work lost.
- The catalogue goes from 1421 translated / 6 untranslated to **1419 / 27**. The extra 21 are the reintroduced strings, awaiting a future translation pass.
- **`--no-fuzzy-matching` on purpose.** A plausible-but-wrong guess at a consent-UI string is worse than showing the English source: a confident mistranslation of *"Material change — ask visitors to consent again"* misleads the administrator about what the button actually does.
- The `.mo` was recompiled and verified byte-identical to a fresh `msgfmt`. That is the file WordPress loads at runtime, and a stale one silently ships the old strings.

Verified the 13 reintroduced msgids are present again (`Footer legal links`, `Pages to link`, `Show legal links in the footer`, `Legal information`, `Decide later`, and the rest of the version-ledger set).

Kept separate from #207 so the contributed translation stays attributable to its author and this realignment reads as a maintenance step rather than an edit to someone else's work. No other locale touched.


fabiodalez-dev:
@coderabbitai full review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev (commented):
## Release review completed

### Findings and corrections
- Compared the Czech catalogue against the current POT, compiled MO output, placeholder formats, plural metadata, and actual visitor-facing render paths.
- Found two translations orphaned by a quote-glyph source change and restored them against the current msgids.
- Translated the visitor-facing legal footer accessibility heading.
- Added a six-locale PO/MO integrity suite that checks headers, format strings, fuzzy entries, source alignment, and exact decoded MO equality with a fresh compile.
- Corrections are in commit f6b0633.

### Validation
- Catalogue integrity suite: 19/19.
- Full unit suite: 62/62.
- msgfmt format/header/domain checks: clean.
- No fuzzy Czech entries.
- Czech coverage after resync: 1422 translated, 24 untranslated.
- PHPStan: clean.

tests/unit/test-i18n-catalogue-integrity-php.php
```diff
@@ -0,0 +1,228 @@
...
+ *   3. Every `.po` has a `.mo` beside it, since a catalogue with no binary is a
+ *      translation that can never load.
+ *
+ * Deliberately NOT asserted: byte-identity between the shipped `.mo` and a
+ * local `msgfmt` run. The `.mo` format carries a hash table whose size and
+ * layout are an implementation detail of the compiling gettext, so byte-compare
+ * turns a contributor's gettext version into a test failure. Comparing decoded
+ * maps pins the property that actually matters and is version-independent.
+ *
+ * Run: php tests/unit/test-i18n-catalogue-integrity-php.php
```
fabiodalez-dev, line 29: **question:** The header documents this as `php tests/unit/test-i18n-catalogue-integrity-php.php`, and nothing in this PR touches `.github/workflows/plugin-check.yml` or `codeql.yml`.

> Is there an existing runner that globs `tests/unit/*.php`, or is this manual-invocation only?

If it's the latter, the suite doesn't yet close the gap its own header describes — a contributor hand-editing a `.po` without recompiling still gets a green CI. Worth adding a step to `plugin-check.yml` (and making sure `gettext` is on the runner image, since `msgfmt` is a hard dependency of assertions 1 and 2).

tests/unit/test-i18n-catalogue-integrity-php.php
```diff
@@ -0,0 +1,228 @@
...
+
+echo "Translation catalogue integrity (.po <-> shipped .mo)\n\n";
+
+$languages = dirname( __DIR__, 2 ) . '/languages';
+$pos       = glob( $languages . '/*.po' );
+sort( $pos );
+
+cat_eq( count( $pos ) > 0, true, 'at least one .po catalogue is present' );
+
+// The whole point is comparing against a compile, so without msgfmt this suite
```
fabiodalez-dev, line 110: **question:** The patch is truncated right here, so I can't see the fallback — but this is the branch that decides whether the suite is real:

- If a missing `msgfmt` results in a **skip**, then on any runner without `gettext` the suite passes unconditionally, which is the same invisible-green failure the file header opens with.
- Similarly, does the script end with `exit( $failed > 0 ? 1 : 0 )`? `$failed` is tracked in `cat_eq()`, but a `php script.php` that prints `FAIL` and exits 0 won't fail a CI step.

My suggestion for the first: treat a missing `msgfmt` as a hard error (non-zero exit with a clear message) rather than a skip. A dependency this suite cannot function without is a setup failure, not an excusable absence.

tests/unit/test-i18n-catalogue-integrity-php.php
```diff
@@ -0,0 +1,228 @@
...
+	}
+	ksort( $map );
+
+	return $map;
+}
+
+echo "Translation catalogue integrity (.po <-> shipped .mo)\n\n";
+
+$languages = dirname( __DIR__, 2 ) . '/languages';
+$pos       = glob( $languages . '/*.po' );
```
fabiodalez-dev: **bug:** `glob()` returns `false` on error, and on PHP 8 both `count( false )` (line 102) and `sort( false )` (line 100) throw a `TypeError` rather than reporting a useful failure. Unlikely in practice since `languages/` is committed, but the guard is free and keeps a missing-directory case reporting as a test failure instead of a fatal.

```suggestion
$pos       = glob( $languages . '/*.po' ) ?: array();
```

fabiodalez-dev (commented):
### Verdict
**question**

### Primary Goal
Realign the Czech catalogue with the current POT via `msgmerge`, restoring the 13 msgids that the 1.25-based contribution in #207 dropped, and recompile the `.mo`.

### Overview
### What this PR does

Three files change:

1. **`languages/faz-cookie-manager-cs_CZ.po`** — `msgmerge`d against the current POT with `--no-fuzzy-matching`. Net effect per the description: 1421→1419 translated, 6→27 untranslated, with the 13 footer-legal-links / policy-version-ledger msgids reintroduced as untranslated rather than absent.
2. **`languages/faz-cookie-manager-cs_CZ.mo`** — recompiled from the above.
3. **`tests/unit/test-i18n-catalogue-integrity-php.php`** — *new*, 228 lines: a standalone harness asserting, per locale, that `msgfmt` accepts the `.po`, that the shipped `.mo` decodes to the same entry map as a fresh compile, and that every `.po` has a `.mo` beside it.

### On the catalogue itself

GitHub did not render patches for the `.po`/`.mo`, so I can't independently verify the 1419/27 counts or the reintroduced msgids — those rest on the description. The reasoning behind `--no-fuzzy-matching` is sound and worth keeping: for consent UI, an English fallback is strictly safer than a confident mistranslation of a destructive-sounding control. Recompiling the `.mo` in the same commit is the right call; a `.po`-only commit is exactly the failure mode the new test is written to catch.

### On the new test

The decoder itself is correct: magic-byte endianness detection, `unpack` offsets, and the choice to key the map on gettext's raw bytes (so `ctxt\x04msgid` and `singular\x00plural` stay distinct) are all right. Comparing *decoded maps* instead of bytes is the correct trade-off and the header comment explains why — though note it reads slightly against the PR description, which cites byte-identity as the verification performed.

My open questions are about whether the harness can actually fire:

- **Nothing in this diff adds it to a workflow.** Only three files change, and neither `plugin-check.yml` nor `codeql.yml` is touched. Unless an existing runner globs `tests/unit/*.php`, this is a manual-invocation script — which means the stale-`.mo` class of bug stays exactly as invisible in CI as the file header says it is today.
- The patch is truncated at the `msgfmt`-availability comment, so I can't see the fallback path or the final exit code. Both matter: a silent skip when `gettext` is missing, or a `$failed` count that never reaches `exit(1)`, turns the suite into a no-op that reports green.

### Concerns

- [ ] Wire the suite into CI (and ensure `gettext` is installed on the runner)
- [ ] Confirm a non-zero exit status when `$failed > 0`
- [ ] Confirm the no-`msgfmt` path fails loudly rather than skipping quietly
- [ ] Guard `glob()` against `false` (one-line, suggestion provided)

### Scope Assessment
The catalogue resync is tightly scoped and touches no other locale, as stated. The 228-line test file is **support, not drift** — it targets precisely the `.po`/`.mo` divergence this PR repairs — but the PR description never mentions it, so a reader expecting a translation-only change gets a repo-wide test harness. Either add a paragraph to the description or split it out; the current mismatch is the only scope issue I see.

### Risk Assessment
Low runtime risk. The `.po`/`.mo` change is data-only and the `--no-fuzzy-matching` policy makes the worst case an untranslated (English) string rather than a wrong one. The real risk is a false sense of safety: if the new suite is never executed by CI, or exits 0 on failure, or skips silently without `msgfmt`, the repository gains a test file but no actual protection — and future contributors will reasonably assume the check is running. The `.mo` decoder validates the index tables but not the individual string offsets, so a deliberately malformed `.mo` yields garbage entries rather than a clean error; that is a robustness nit, not a security issue, since the input is a repo-controlled file.

### Reuse Notes
- `.github/workflows/plugin-check.yml` and `.github/workflows/codeql.yml` already exist — the new suite should hook into the former (or whatever step runs `tests/unit/`) rather than remain manual.
- Other files under `tests/unit/` presumably share a helper/assertion convention; `cat_eq()` may duplicate an existing assertion helper worth reusing instead of redeclaring.

### Action Items
- [ ] Add the new suite to a GitHub workflow and install `gettext` on the runner, otherwise the check never runs.
- [ ] Confirm the script ends with a non-zero exit when `$failed > 0` (truncated in the diff I was given).
- [ ] Make the missing-`msgfmt` path fail or hard-error rather than skip silently — a quiet skip reproduces the exact blind spot the file header describes.
- [ ] Apply the `glob()` false-guard suggestion.
- [ ] Mention the new test file in the PR description, or split it into its own PR.

### Separate PR Suggestions
- `tests/unit/test-i18n-catalogue-integrity-php.php` plus its CI wiring could stand as its own PR — it's a repo-wide, all-locales guarantee rather than a Czech-catalogue change, and separating it would let the `.po`/`.mo` resync merge on its own merits. Keeping them together is defensible given the test targets exactly this failure mode; if it stays, please add a line about it to the PR description, which currently reads as translation-only.

tests/unit/test-i18n-catalogue-integrity-php.php
```diff
@@ -0,0 +1,228 @@
...
+	} else {
+		return 'bad magic (not a .mo file)';
+	}
+
+	$header = unpack( $fmt . '5', substr( $blob, 4, 20 ) );
+	$count  = $header[2];
+	$o_orig = $header[3];
+	$o_tran = $header[4];
+
+	if ( $o_orig + ( 8 * $count ) > strlen( $blob ) || $o_tran + ( 8 * $count ) > strlen( $blob ) ) {
```
fabiodalez-dev, line 79: **suggestion:** This validates that the two **index tables** fit inside the blob, but the per-entry offsets read at lines 87–88 (`$o[2]`, `$t[2]`) are never bounds-checked. On a truncated or corrupt `.mo`, `substr()` returns a short string or `""` instead of erroring, so the map silently decodes to wrong values and the failure surfaces as a confusing entry mismatch rather than `'string tables run past end of file'`.

Since the whole point of this decoder is to be the authority on what shipped, a check inside the loop — offset + length must not exceed `strlen( $blob )`, else return a descriptive error — would make a corrupt binary report as corrupt. Repo-controlled input, so this is robustness rather than security.

tests/unit/test-i18n-catalogue-integrity-php.php
```diff
@@ -0,0 +1,228 @@
...
+ * turns a contributor's gettext version into a test failure. Comparing decoded
+ * maps pins the property that actually matters and is version-independent.
+ *
+ * Run: php tests/unit/test-i18n-catalogue-integrity-php.php
+ */
+
+$tests_run = 0;
+$failed    = 0;
+
+function cat_eq( $actual, $expected, $label ) {
```
fabiodalez-dev, line 35: **style:** Minor consistency point: `faz_decode_mo()` carries the plugin prefix but `cat_eq()` doesn't. In a WordPress codebase (and under WPCS prefix rules) an unprefixed global function name like `cat_eq` is the kind of thing that trips a sniff or collides the day someone runs two of these standalone scripts through a shared bootstrap. `faz_cat_eq()` would match the neighbouring helper — not blocking, and I'd skip it if the other files in `tests/unit/` already use unprefixed helpers, since matching the local convention wins.

PR #210, opened 2026-08-03
perf: reduce per-request overhead, autoload bloat, and unbounded queries

fabiodalez-dev:
Performance audit (docs/performance-audit-2026-08-03.md) across DB queries,
options autoload, asset loading and hot-path hooks. Fixes applied:

Options autoload:
- faz_banner_template (+ language variants) no longer autoloaded (~30-45 KB
  of rendered HTML+CSS per banner x language was in alloptions on every
  request); one-time migration demotes existing rows
- scanner bookkeeping, admin-notice map, onboarding and migration flags now
  written with autoload=false

Hot path:
- memoize Gcm_Settings::get() (was 5x option read + recursive sanitize/request)
- memoize Geolocation database path and per-IP country/region lookups
- schedule_cleanup() no longer runs on every frontend request
- scanner cron hooks register lazily (no controller instantiation per request)
- static regex cache in output-buffer attribute extraction
- admin cookie-definitions notice no longer decodes the 2.8 MB bundled JSON
  (metadata cached in an option keyed by file mtime+size)

Database:
- consent-log CSV export streams in keyset-paginated batches (was unbounded
  SELECT * of the highest-volume table)
- retention DELETEs batched with LIMIT 1000 (consent logs + pageviews)
- pageviews retention cleanup wired into faz_daily_cleanup (was dead code;
  table grew forever)
- new indexes: faz_consent_logs.banner_slug, faz_pageviews(event_type,created_at)
- [faz_cookie_table] shortcode reads through the cache layer (was 2 uncached
  full-table scans per public render)
- scanner discovery loop invalidates caches once per batch instead of per
  cookie (Base_Controller bulk mode)
- has_country_dependent_banners() classifies raw rows instead of hydrating a
  full Banner per row per request

Assets:
- minified builds for faz-dsar, faz-dnsmpi, wca, microsoft-consent (the first
  two load on every page by design); added to build:min pipeline
- cookie-policy shortcode uses its own style handle (collision with the
  generator's file-backed handle silently dropped one registration)

Tests: 61/61 unit suites pass; geo suites updated for the new per-request
path memoization.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_016SUoBNdVoDL4yKacLC6QsX

## Summary by CodeRabbit

* **Prestazioni**
  * Ottimizzato il caricamento degli asset frontend, con supporto a file esterni versionati, cache migliorata e caricamento differito.
  * Ridotto il consumo di memoria durante esportazioni, pulizie e operazioni in blocco.
  * Velocizzati rilevamento dei consensi, geolocalizzazione e gestione dei cookie.

* **Bug Fix**
  * Corretti possibili conflitti negli stili degli shortcode.
  * Migliorata la gestione degli asset JavaScript minificati.
  * Rafforzata la pulizia dei dati durante disinstallazione e manutenzione automatica.

* **Documentazione**
  * Aggiunto un audit delle prestazioni con risultati, verifiche e raccomandazioni future.


fabiodalez-dev:
Superseded by #212 and #213, which carry this work split in two so the front-end hot path and the data layer can be reviewed separately.

The split also isolated the part that matters most: the `<noscript>` gating change in here is a **compliance fix, not a performance one** — visitors without JavaScript were loading tracking pixels before any consent, and they cannot consent at all since the banner is JavaScript. It now sits in its own commit in #212 rather than buried in a 28-file performance branch.

Other differences from this branch:

- Commit messages rewritten. The originals carried co-author trailers and session links that do not belong in this repository's history. No code changed in the rewrite.
- Review findings addressed in both halves: atomic asset writes, the reaper no longer deleting files that full-page caches still reference, per-blog directory resolution on multisite, a failed `filemtime()` no longer read as "old", the cache flush moved into `finally` so a mid-loop failure cannot leave rows behind a stale cache, and the `<noscript>` rewrite no longer neutralising whitelisted or unrelated resources.
- The scanner flush-on-error change that shared a commit with front-end work is in neither branch yet; worth raising separately if still wanted.

Closing to avoid merging the same work twice — and with the attribution the split removed.

PR #215, opened 2026-08-07
Translate cookie duration and description per language, not just categories

### LINKED ISSUE #214: Automatically translate individual cookies (duration and description), not just categories
Hi, congratulations on this very complete plugin. Would it be possible to add this feature?

**Problem / Context**
Currently, when the cookie banner is configured in Italian and English, automatic translation only covers the cookie categories. Individual cookies within each category — specifically the "duration" and "description" fields — remain in a single language regardless of the language selected in the banner.

**Current behavior**

Banner in Italian → categories are translated, but the duration/description of individual cookies stay in English (or whatever language they were originally entered in).

**Expected behavior**

Banner in Italian → the "duration" and "description" fields of individual cookies should also be shown in Italian.
Banner in English → the same fields should be shown in English.
Translation should follow the same logic already used for categories.

**Why this is useful**
It improves the consistency and professionalism of the banner: a user who selects Italian expects all the content — not just the category labels — to be in Italian. Currently the banner looks "mixed" and is less clear for the end user.

**Practical example**
Category "Statistics" (translated to IT) → cookie _ga, duration "2 years" (not translated) instead of "2 anni".

fabiodalez-dev:
Closes #214.

With the banner configured in Italian and English, only the category labels were translated. The duration and description of each cookie stayed in whatever language they were entered in, so the banner read as a mix — `Statistiche` as the category, `2 years` as the cookie duration underneath it.

`Cookie::get_translations()` was a stub returning `''`. It now resolves through a bundled catalogue: curated descriptions keyed by cookie slug (with prefix entries such as `_ga_` covering the concrete names generated at runtime), and a unit catalogue that translates simple retention periods like `2 years` → `2 anni`.

### The fallback never overwrites anything

Every caller consults it **only** when the stored value for the requested language is empty, so an admin's own wording always wins. There is no database rewrite and no schema change: legacy rows gain translations on read. Free-form or ambiguous durations (`1 year or longer`) resolve to nothing and the stored value is preserved.

Wired into the three surfaces that render cookie inventory: the `Cookie` model, the `[faz_cookie_table]` shortcode, and the Cookie Policy renderer. The upgrade migration rotates the prepared cookie/category caches and the language-keyed policy fragments, since those hold pre-translation payloads even though nothing in the database changed.

### Two things found while verifying, fixed here

**Croatian was missing.** `hr_HR` ships as a plugin locale but had no entry in the duration catalogue, so a Croatian banner silently kept English durations. Nothing failed — the fallback is deliberately non-destructive, which is exactly what makes a missing language invisible. Added with its plural rule (few = 2–4, excluding the 12–14 teens). The part that matters is the guard: a test comparing `languages/*.po` against the catalogue keys, so the next locale cannot be added without one.

**Capitalisation was being dropped.** Inventories store `Session` far more often than `session`, while the catalogue is lowercase so it can also sit mid-phrase. The duration column therefore read `Session` in English and `sessione` in Italian — the mixed-language look this issue is about, reintroduced by its own fix. The translation now carries the source's leading capitalisation; numeric durations start with a digit and are untouched.

### Verification

- 31 assertions in `tests/unit/test-cookie-content-i18n-php.php`: plural forms across Italian, Czech, Polish and Croatian, locale normalisation (`it_IT` → `it`), prefix descriptions, free-form durations left alone, legacy rows, the renderer path, the cache-rotation migration, and the locale-coverage guard.
- Checked on a live install, not only in unit: with Italian active, `en="Session"` renders as `it="Sessione"` across the inventory.
- No N+1. The shortcode builds a `Cookie` per row, so I measured it rather than reasoning about it: 40 rows, 0 additional queries — the rows carry `cookie_id`, which takes the `read_direct` path.
- Unit 62/62, compliance 121/121, verify 14/14, and 78/78 on the specs this can touch (cookie policy, cookie table, multilingual, banner i18n).

### Known gap

Descriptions exist for `en` and `it` only (67 entries each). For German, French, Dutch and Croatian the durations now translate but the descriptions do not. That is not a defect — the fallback preserves whatever is stored — but the coverage is asymmetric and worth deciding on separately.

## Summary by CodeRabbit

* **Nuove funzionalità**
  * Localizzate descrizioni e durate dei cookie in base alla lingua del visitatore nel banner, nello shortcode `[faz_cookie_table]` e nella Cookie Policy.
  * Aggiunti cataloghi di traduzione con fallback per valori multilingua, descrizioni personalizzate e durate riconosciute.
  * Introdotti filtri per personalizzare o disabilitare le traduzioni dei contenuti.

* **Correzioni**
  * Preservati i valori personalizzati e quelli non riconosciuti durante la localizzazione.
  * Migliorata l’invalidazione delle cache dopo le migrazioni per mantenere aggiornati i contenuti localizzati.


fabiodalez-dev:
All four addressed, with one correction to the review itself.

Applied as described: the empty-source-duration guards, running the cache rotation on the front-end upgrade path (`install()` off `check_version` on `init`, since `set_object_cache()` writes with no expiry and would not self-heal), and clearing policy fragments beyond the selected languages — `resolve_lang()` takes the language from the shortcode attribute, the page default or the WP locale, so it is not bound to `faz_selected_languages()` at all.

The plural rule is right for Croatian and wrong for Polish. The bundled catalogue shows it directly:

```
hr.hour  one=sat  other=sati   CLDR: 21 is one   -> "21 sat"
pl.year  one=rok  other=lat    CLDR: 21 is many  -> "21 lat"
```

CLDR gives Polish `one` for exactly 1 and nothing else; Croatian gives it to every number ending in 1 except 11. Applying the suggestion verbatim would render "21 rok". Applied to Croatian only.

Both directions are now pinned: removing the Croatian rule fails "21 sat", and extending it to Polish fails "21 lat", so the suite catches the suggestion as written. 36 assertions, and I checked each new one detects the absence of its fix — two of them did not at first, which is why the cases exist.

fabiodalez-dev (commented):
## Release review completed

### Findings and corrections
- Reviewed the cookie content data flow from catalogue lookup through policy rendering and shortcode output, including default-language and translated-language behavior.
- Found a backward-compatibility regression: a customized or legacy description could be overwritten by the stock catalogue description in the default language.
- Corrected in commit 0dc2ffa by preserving explicit legacy/custom content precedence in both rendering paths while retaining translated catalogue content where no customization exists.
- Added three regression cases covering renderer and shortcode behavior across custom, legacy, and stock values.

### Validation
- Targeted suites: 39/39.
- Full unit suite: 62/62.
- Cookie-content end-to-end matrix: 50/50.
- Compliance suite: 121/121.
- Final verification: 14/14.
- PHPStan: clean.
- Cookie-duration data validated as JSON.

fabiodalez-dev (commented):
### Verdict
**looks good to me**

### Primary Goal
Translate each cookie's duration and description into the visitor's language via a bundled, non-destructive fallback catalogue, instead of translating only the category labels.

### Overview
### TL;DR

Solid, well-scoped fix for #214: `Cookie::get_translations()` stops being a stub and resolves through a bundled catalogue (per-locale cookie descriptions + a duration-unit catalogue), applied **only** when the stored value for the requested language is empty. Additive, read-side, no schema change — I'd merge this once the two documentation/consistency points below are settled.

---

### What this PR does

Three surfaces render cookie inventory, and all three now consult the same fallback:

1. **`Cookie` model** (`admin/modules/cookies/includes/class-cookie.php`) — the canonical read path.
2. **`[faz_cookie_table]` shortcode** (`includes/class-cookie-table-shortcode.php`).
3. **Cookie Policy renderer** (`admin/modules/cookie-policy-generator/includes/class-renderer.php`).

The new logic lives in `includes/class-cookie-content-i18n.php`, backed by data files:

- `admin/modules/cookies/includes/contents/cookies/<locale>.json` — per-cookie descriptions, keyed by cookie slug, with **prefix keys** (`_ga_`, `wp-settings-`, `comment_author_`) covering runtime-generated names.
- `admin/modules/cookies/includes/contents/duration-units.json` — retention-period units with plural rules per locale.

`includes/class-activator.php` / `includes/class-store.php` rotate the prepared cookie/category caches and the language-keyed policy fragments on upgrade, because those hold pre-translation payloads even though nothing in the DB changed. That's the right call — a purely read-side change still invalidates anything that memoised the read.

### What I like

- **Non-destructive by construction.** The fallback is consulted *only* when the stored value for the target language is empty, so an admin's own wording always wins, and unrecognised durations (`1 year or longer`) fall through untouched. This is the property that makes the whole feature safe to ship without a migration.
- **Prefix keys are the right data model.** `_ga_ABC123` and `wp-settings-1` carry runtime suffixes; an exact-match table would never hit them. Longest-prefix-first resolution is what makes the catalogue useful in practice.
- **The locale-coverage guard is the best part of this PR.** Croatian was silently keeping English durations precisely *because* the fallback is non-destructive — a missing locale fails invisibly. Adding a test that diffs `languages/*.po` against the catalogue keys converts that class of bug from "discovered by a user" to "caught in CI." That's the durable fix; the `hr` entry itself is just the symptom.
- **Capitalisation carry-over.** Catching that `Session` → `sessione` reintroduced the exact mixed-language look the issue complains about, and fixing it by carrying the source's leading case (numerics untouched), is a nice piece of self-review.
- **Performance claim is measured, not asserted.** "40 rows, 0 additional queries — the rows carry `cookie_id`, which takes the `read_direct` path" is exactly the level of evidence an N+1 concern deserves.

### Points to settle before merge

**1. The stated known gap doesn't match what shipped.** The description says descriptions exist for `en` and `it` only, but the diff adds `bg.json`, `cs.json`, `de.json`, `es.json`, `fi.json`, `fr.json`, `hr.json`, `hu.json`, `nl.json`, `pl.json`, `pt.json`, `pt-br.json` — several of them with full description sets. Either the gap note is stale, or those files aren't wired into the resolver yet. Not a defect either way, but a future reader will trip on it. Please reconcile the description (and `readme.txt` / `CHANGELOG.md` if they repeat the claim).

**2. Two lookup-side invariants worth confirming with a test each** (the JSON is internally consistent; these are resolver concerns):

- **Case.** Catalogue keys are lowercase (`ide`, `dsid`, `muid`) while the real cookies are `IDE`, `DSID`, `MUID`. This only works if the lookup lowercases the incoming cookie name. A one-line assertion on `IDE` would pin it.
- **Prefix collision order.** `wordpress_` is a prefix of both `wordpress_logged_in_` and `wordpress_test_cookie`. If resolution isn't strictly longest-prefix-first, every logged-in cookie silently gets the generic admin-auth text — again, a *wrong but plausible* result, which is the hardest kind to notice. The PR mentions prefix descriptions are tested; an explicit longest-wins case would make the invariant explicit rather than incidental.

**3. Key-set parity across catalogue files.** `bg.json` and `cs.json` have identical key sets in identical order — a good invariant. Consider extending the existing coverage guard to assert *key-set parity across all catalogue files*, so a locale can't silently drift a few entries behind as the catalogue grows. Same reasoning as the `.po` guard: make the invisible failure visible.

### Scope and risk

Everything in the diff maps to the goal. The two files that could be called "beyond the ask" — `class-activator.php` and `class-store.php` — are cache rotation, which is a genuine consequence of changing what a cached read returns, not drift. No auth, schema, or public-API surface touched.

Residual risk is low and mostly cosmetic: a mis-keyed catalogue entry shows the stored value (status quo), and a bad prefix match shows a plausible-but-wrong description. The one real hazard is JSON that fails to parse — worth confirming the loader degrades to the stored value rather than fataling if a catalogue file is ever malformed or truncated by a partial deploy.

### Suggested follow-up (separate PR)

The asymmetric coverage — durations translated for all locales, descriptions only for some — is a product decision, not a code defect. Worth its own PR so the translation work can land locale-by-locale without re-reviewing the resolver.

### Scope Assessment
On-scope. All changes trace to the stated goal: the i18n resolver and its data files are CORE; the three render surfaces (`Cookie`, shortcode, policy renderer) are the required call sites; `class-activator.php` and `class-store.php` are SUPPORT — cache rotation is a real consequence of changing what a cached read returns, not drift. No auth, DB schema, or public API touched. `CHANGELOG.md` / `readme.txt` are documentation. No unrelated reformatting or renames observed.

### Risk Assessment
Low. The fallback is consulted only when the stored value for the requested language is empty, so no admin-authored text can be overwritten and no DB rows change — worst case is a cosmetic regression to the current behaviour. Three residual risks: (1) case sensitivity — catalogue keys are lowercase (`ide`, `dsid`, `muid`) while real cookies are `IDE`, `DSID`, `MUID`, so the lookup must normalise; (2) prefix-collision order — `wordpress_` prefixes both `wordpress_logged_in_` and `wordpress_test_cookie`, and anything other than longest-prefix-first silently yields a plausible-but-wrong description; (3) a malformed or partially deployed catalogue JSON should degrade to the stored value rather than fatal. All three are resolver-side and testable in one assertion each. No security surface: the catalogue is bundled static data, not user input, and the change is read-only.

### Reuse Notes
- The locale-coverage guard (`languages/*.po` vs catalogue keys) is the reusable idea here — extend the same pattern to assert key-set parity across all `contents/cookies/*.json` files so a locale can't silently drift behind.
- The prefix-key convention (`_ga_`, `wp-settings-`, `comment_author_`) is now established data-side; future catalogue additions should follow it rather than enumerating concrete runtime cookie names.
- Cache rotation in `class-activator.php` is the established place for invalidating prepared payloads on upgrade — reuse it for any future read-side transformation rather than adding a parallel invalidation path.

### Action Items
- [ ] Reconcile the 'Known gap' note with what actually shipped: the diff adds 12 locale catalogues, not just `en`/`it`. Update the PR description (and `readme.txt`/`CHANGELOG.md` if they repeat it), or state which files are not yet wired into the resolver.
- [ ] Add an assertion that lookup is case-insensitive — e.g. cookie `IDE` resolves against the lowercase `ide` key — so the catalogue's lowercase convention is pinned rather than assumed.
- [ ] Add an explicit longest-prefix-wins assertion for `wordpress_logged_in_...` vs the shorter `wordpress_` key, so prefix-collision ordering is a tested invariant rather than incidental behaviour.
- [ ] Extend the coverage guard to assert key-set parity across all `contents/cookies/*.json` catalogues, catching a locale that drifts entries behind as the catalogue grows.
- [ ] Confirm the catalogue loader degrades to the stored value (rather than fataling) if a JSON file is malformed or truncated by a partial deploy.

### Separate PR Suggestions
- If `es.json` and `fi.json` are not deliberate scope for #214, extract both locale catalogues into their own PR — one that also adds the matching duration-catalogue entries and extends the `languages/*.po` coverage guard to description catalogues, so a locale can never ship descriptions without durations again.
- Estendere il catalogo delle descrizioni oltre `en`/`it` (de, fr, nl, hr) — il gap noto dichiarato nella descrizione della PR, meglio affrontato separatamente dato che è puro contenuto e richiede revisione linguistica.
- Curated descriptions currently exist for `en` and `it` only (67 entries each), so German, French, Dutch and Croatian get translated durations but untranslated descriptions. The PR body flags this as a deliberate known gap — filling out the description catalogue for the remaining bundled locales, with the same `languages/*.po`-vs-catalogue coverage guard extended to descriptions, is its own change.
- If the `'{' === $value[0]` sniff also exists in `decode_i18n_text()` and elsewhere in the codebase, normalising multilingual-payload detection into one shared helper (trim + `json_decode` + shape check) would be a worthwhile follow-up, kept out of this PR so the translation change stays reviewable in isolation.
- Complete the description catalogues for the locales that currently translate durations but not descriptions (de, fr, nl, hr and any others). The asymmetric coverage is a product decision, not a defect, and keeping it separate lets translation work land locale-by-locale without re-reviewing the resolver.

admin/modules/cookies/includes/contents/cookies/de.json
```diff
@@ -0,0 +1,203 @@
...
+        "description": "Google-Analytics-Cookie (Classic), das die Traffic-Quelle oder Kampagne speichert."
+    },
+    "__utmt": {
+        "description": "Google-Analytics-Cookie (Classic) zur Begrenzung der Anfragerate."
+    },
+    "_gcl_au": {
+        "description": "Cookie des Google-Ads-Conversion-Linkers."
+    },
+    "ide": {
+        "description": "DoubleClick-/Google-Cookie für zielgerichtete Werbung."
```
fabiodalez-dev, line 72: **bug:** **`domains` scoping is missing here but present in `en.json`.**

In `en.json`, `ide`, `dsid` and `fr` each carry a `domains` array (`doubleclick.net`, `facebook.com`). In `de.json` the same three keys have only a `description`.

> If the resolver reads `domains` from the language file rather than from a single base catalogue, a German site loses the guard on exactly the keys where it matters most.

`fr` and `ide` are three-letter, entirely generic names — a site's own framework or a first-party session cookie can plausibly be called `fr`. On an English banner the `domains` scope prevents the "Facebook advertising cookie" copy from being attached to it; on a German banner it wouldn't. In a Cookie Policy that's a factually wrong disclosure, not just a cosmetic slip.

If `domains` is only ever read from `en.json` as the base, this is a non-issue and the suggestion below is a harmless consistency fix — but it's worth confirming which of the two it is.

```suggestion
        "description": "DoubleClick-/Google-Cookie für zielgerichtete Werbung.",
        "domains": [
            "doubleclick.net"
        ]
```

Same treatment needed for `dsid` (`doubleclick.net`) and `fr` (`facebook.com`) further down.

```suggestion
"description": "DoubleClick-/Google-Cookie für zielgerichtete Werbung.",
        "domains": [
            "doubleclick.net"
        ]
```

admin/modules/cookies/includes/contents/cookies/de.json
```diff
@@ -0,0 +1,203 @@
+{
```
fabiodalez-dev, line 1: **question:** **Is `de.json` actually wired into the resolver?**

The PR description states:

> Descriptions exist for `en` and `it` only (67 entries each). For German, French, Dutch and Croatian the durations now translate but the descriptions do not.

But this file is a complete German description catalogue. Either the description is stale, or this file ships without being read — in which case it's dead weight that will drift out of sync with `en.json` over time.

Related: `en.json` is **231** lines against **203** here. The three visible `domains` blocks in `en.json` account for ~9 of those lines, leaving roughly six entries that may be missing from the German catalogue.

A missing key degrades exactly the way the Croatian duration gap did — silently, to the stored English string, producing the mixed-language banner that #214 is about. The PR already adds a locale-coverage guard for the duration catalogue for precisely this reason; the description catalogues would benefit from the same shape:

```php
// every key in en.json must exist in every other catalogue
foreach ( glob( $dir . '/*.json' ) as $file ) {
    $missing = array_diff_key( $en, json_decode( file_get_contents( $file ), true ) );
    $this->assertSame( [], $missing, basename( $file ) . ' is missing keys' );
}
```

That way the next language can't be added half-populated without the suite noticing.

admin/modules/cookies/includes/contents/cookies/de.json
```diff
@@ -0,0 +1,203 @@
...
+        "description": "Passt die Administrationsoberfläche für jeden Nutzer an."
+    },
+    "wordpress_": {
+        "description": "WordPress-Authentifizierungscookie für den Administrationsbereich."
+    },
+    "wp_lang": {
+        "description": "Speichert die bei der Anmeldung gewählte Sprache."
+    },
+    "comment_author_email_": {
+        "description": "Speichert die E-Mail-Adresse des Kommentierenden zur Erleichterung."
```
fabiodalez-dev, line 21: **style:** Minor copy note, batched for the three `comment_author_*` entries (lines 21, 24, 27): **"zur Erleichterung"** reads as a literal rendering of "for convenience" and is unidiomatic in German — it suggests relief/alleviation rather than convenience.

Since this text lands in a Cookie Policy, which is a compliance document a visitor may actually read, natural phrasing is worth the two-word change. The WordPress behaviour being described is prefilling the comment form on subsequent visits, so naming that is both more idiomatic and more accurate:

```suggestion
        "description": "Speichert die E-Mail-Adresse des Kommentierenden für zukünftige Kommentare."
```

Same for `comment_author_url_` ("… die Website-Adresse … für zukünftige Kommentare.") and `comment_author_` ("… den Namen … für zukünftige Kommentare.").

```suggestion
"description": "Speichert die E-Mail-Adresse des Kommentierenden für zukünftige Kommentare."
```

admin/modules/cookies/includes/contents/cookies/es.json
```diff
@@ -0,0 +1,203 @@
+{
```
fabiodalez-dev, line 1: **question:** **Are these two catalogues intentionally part of this PR?**

The PR description is explicit:

> Descriptions exist for `en` and `it` only (67 entries each). For German, French, Dutch and Croatian the durations now translate but the descriptions do not.

But this pass contains `es.json` and `fi.json` — full Spanish and Finnish **description** catalogues, 203 lines each. Spanish and Finnish are not even mentioned in the gap list.

Two readings, and they lead to different actions:

1. **Intentional** → the "Known gap" paragraph is stale and should be updated before merge, otherwise the next reader will believe description coverage is `en`/`it` when it is `en`/`it`/`es`/`fi`.
2. **Drift** → these came from a parallel branch and should be extracted into their own PR, where the locale-coverage question gets its own review.

The follow-up that matters either way:

> Do `es` and `fi` also have entries in the **duration** catalogue?

If they don't, this reintroduces the exact symptom issue #214 reports, mirrored: the banner would read `Cookie de Google Analytics utilizada para distinguir a los usuarios.` above a duration still rendered as `2 years`. The Croatian finding in your own description proves the failure mode is silent — the fallback is non-destructive by design, so a half-covered locale looks fine in tests and wrong in production.

Related: the coverage guard you added compares `languages/*.po` against the **duration** catalogue keys. Does it also assert description-catalogue coverage? If not, `es.json` and `fi.json` landing here is precisely the case it wouldn't catch.

admin/modules/cookies/includes/contents/cookies/es.json
```diff
@@ -0,0 +1,203 @@
...
+    },
+    "_grecaptcha": {
+        "description": "Utilizada por Google reCAPTCHA para distinguir entre personas y bots."
+    },
+    "fazcookie-consent": {
+        "description": "Preferencias de consentimiento de cookies establecidas por el visitante."
+    },
+    "_clck": {
+        "description": "Cookie de Microsoft Clarity para análisis."
+    },
```
fabiodalez-dev, line 100: **suggestion:** **`__cfduid` no longer exists.**

Cloudflare [retired `__cfduid` in May 2021](https://blog.cloudflare.com/deprecating-cfduid-cookie/) and replaced its role with `__cf_bm`, which both catalogues already cover on the line above.

Not a bug — a scanner will simply never match it, so the entry is inert. But a cookie policy is a compliance document, and listing a cookie the site cannot possibly set is the kind of thing an auditor flags. If the entry is kept deliberately (to describe legacy inventories imported from older scans, which is a legitimate reason), a short note in the source catalogue explaining why would save the next maintainer the same lookup.

Worth checking whether `en.json` / `it.json` carry the same entry — if the catalogues are generated from one source, this should be decided once at the source rather than per-locale.

PR #219, opened 2026-08-11
test(policy): make the ledger-seeding test establish the state it claims to test

fabiodalez-dev:
`policy-revision-link.spec.ts` opens with a test called *"first load seeds the ledger silently — the feature itself never prompts"* that never arranged a first load. The file's `beforeAll` only snapshots the policy settings, so this test — the first in the file — inherited whatever ledger state the site happened to be carrying, and asserted only that no notice was on screen.

That left it wrong in both directions.

**It failed for no defect.** A site holding a genuine unacknowledged policy change shows the notice *correctly*, and the test read that correct behaviour as a failure. That is exactly what happened here: a half-finished suite run left the ledger acknowledged at an older hash, and the test went red on `main` and on every branch alike — pointing at the feature instead of at the leftover state. I only established it was environmental by deploying `main` and finding the identical hash and the identical `changed` status there.

**And it passed while blind.** `Version_Ledger::evaluate()` returns `'seeded'` whether or not it actually records anything, so deleting the `acknowledge()` call inside it — breaking seeding completely — still renders no notice, and the old assertion still passed. A test named after seeding could not distinguish working seeding from absent seeding.

## What changed

The test now establishes its precondition instead of hoping for it: it clears the ledger through WP-CLI so the page really is meeting an unacknowledged install, then asserts **both** halves of the claim — that no notice appeared, and that the ledger now holds the current hash. A second load must still be silent and must not have quietly re-seeded to a different value.

The ledger is read through WP-CLI rather than through the admin page on purpose. The page is the thing under test; asking it whether seeding happened would only ever return its own opinion.

## Verification

A test-hygiene change that is itself unverified is just a different guess, so this was checked in three directions:

| scenario | result |
|---|---|
| clean ledger | 4/4 green |
| the `changed` state that used to break it | green — the point of the change |
| seeding removed from `Version_Ledger::evaluate()` | red on the new assertion (`Received string: ""`), where the old test passed |

`tests/unit/test-version-ledger.php` — 47/47, untouched.

## Notes

Skips when `WP_PATH` is absent, matching the other WP-CLI-dependent specs: without it the precondition cannot be established, and a test that cannot arrange its own state should say so rather than run on someone else's.

No product code is touched — this is a test-only change, branched from `main`, independent of the open PRs.

## Summary by CodeRabbit

* **Test**
  * Rafforzata la verifica del primo accesso alla pagina delle policy.
  * Verificata la generazione di un identificativo di revisione completo e valido.
  * Confermata la persistenza dello stesso identificativo nelle aperture successive, senza notifiche indesiderate.
  * Aggiunti controlli sullo stato iniziale e sulla sua pulizia, garantendo risultati affidabili e riproducibili.
  * Gestita l’assenza dell’ambiente necessario per eseguire correttamente il test di primo caricamento.

### DIFF
--- a/tests/e2e/specs/policy-revision-link.spec.ts
+++ b/tests/e2e/specs/policy-revision-link.spec.ts
@@ -17,10 +17,31 @@ import { test, expect, completeAdminLogin } from '../fixtures/wp-fixture';
 import type { Browser, BrowserContext, Page } from '@playwright/test';
 import { getAdminNonce, fazApiGet, fazApiPost } from '../utils/faz-api';
 import { clickFirstVisible } from '../utils/ui';
+import { WP_PATH, wpEval } from '../utils/wp-env';
 
 const ADMIN_PAGE = '/wp-admin/admin.php?page=faz-cookie-manager-cookie-policy';
 const NOTICE = '#faz-cp-version-notice';
 const HASH_PAIR_RE = /[0-9a-f]{6}\.[0-9a-f]{6}\s*→\s*[0-9a-f]{6}\.[0-9a-f]{6}/;
+const HASH_RE = /^[0-9a-f]{6}\.[0-9a-f]{6}$/;
+
+// Version_Ledger::OPTION / ::DOCUMENT. Read through wp-cli rather than through
+// the admin page, because the page is the thing under test: asking it whether
+// seeding happened would only ever return its own opinion.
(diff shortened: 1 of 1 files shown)


fabiodalez-dev (commented):
### Overview
I'm being asked to produce a code review as JSON. Let me analyze the diff.

PR #220, opened 2026-08-11
fix(blocking): stop blocking Elementor's own lightbox helper

### LINKED ISSUE #218: Elementor Lightbox blocked
Hi,

currently Elementor lightboxes won't show up until you accept all cookies because `elementor/assets/lib/share-link` is a blocked functional cookie by default. If you decline the cookies it won't load the script and blocks the lightbox.
It's an internal helper script, not a third-party tracker and shouldn't be flaged/blocked at all. I've added it to the white-list and it works fine.

fabiodalez-dev:
Fixes #218.

Elementor lightboxes did not open until a visitor accepted everything. `elementor/assets/lib/share-link` shipped in the provider database as a `functional` pattern, so declining consent blocked the script the lightbox depends on — a core page-builder feature broken for everyone who said no, with the reporter having to find the whitelist themselves to get their site working.

## Why this is a removal, not a narrowing

The entry was wrong on its own terms:

- The pattern addresses a **first-party** file served from the site's own plugins directory. Nothing leaves the site when it loads.
- It declared **zero cookies**. What consent governs is storage and third-party transmission; a script that does neither has nothing to consent to.
- `share-link` builds share URLs. The network is contacted only if a visitor clicks a share button — their own navigation, not a background call.

So the block bought no privacy and cost a working site. The usual asymmetry, that over-blocking is the safe error, only holds when there *is* something to block. Here it was pure cost.

The blocker template goes with it. Its description promised to block "Elementor embeds, widgets, or third-party content" while its only pattern was that helper, so an admin who clicked it got a broken lightbox and no third party blocked. The third parties Elementor can actually embed — YouTube, Vimeo, Google Maps, reCAPTCHA — are covered by their own provider entries and are unaffected.

## Existing installs

Anyone who copied the pattern into their custom blocking rules keeps it. Those are admin-authored and this must not silently rewrite them; removing it there is one click.

## Tests

`tests/unit/test-elementor-not-blocked-php.php` pins that no provider and no template names an Elementor asset path, and matches the reported script URL against every pattern **the way the blocker does** — substring — so a re-introduction under a different spelling is caught too. It also asserts the database is still 300+ entries, so the checks cannot pass by the file having been emptied.

Verified non-vacuous: restoring the provider entry fails two of the five checks. 62/62 unit suites.

## Scope note

Deliberately not generalised into "no provider may declare zero cookies", though that rule would have caught this. A dozen legitimate entries share the shape because they match the script *handle* of a plugin that then loads a real third party (Site Kit, HubSpot, the Facebook pixel). Removing blocking is the dangerous direction, so the guard stays narrow and evidence-led.

## Summary by CodeRabbit

* **Modifiche**
  * Rimosso il supporto al blocco dei cookie per Elementor.
  * Eliminato Elementor dall’elenco dei provider riconosciuti.


includes/data/known-providers.json
```diff
@@ -1515,14 +1515,6 @@
             "__kla_id"
```
fabiodalez-dev, line 1515: **question:** Removing the entry here fixes **fresh** installs cleanly. What I can't tell from the diff is whether it fixes the sites that are already broken.

> Is `known-providers.json` read at runtime on every request, or is it seeded into an option / scanned-cookies table at activation or on the first scan?

If it's read live, this is complete and nothing more is needed. If it's seeded, every existing install keeps the stale `elementor/assets/lib/share-link` pattern in its stored rules and stays broken — which is precisely the reporter's situation in #218 — so an upgrade routine would be needed to strip it from **shipped** rules while leaving admin-authored custom rules untouched, as the PR description already commits to.

Worth stating the answer explicitly in the PR body either way, since "the JSON no longer ships it" and "the visitor's lightbox works again after updating" are only the same statement in the first case.

tests/unit/test-elementor-not-blocked-php.php
```diff
@@ -0,0 +1,105 @@
...
+	$failed++;
+	echo "  \033[31mFAIL\033[0m {$label}\n";
+	echo "        expected: " . var_export( $expected, true ) . "\n";
+	echo "        actual:   " . var_export( $actual, true ) . "\n";
+}
+
+$root      = dirname( __DIR__, 2 );
+$providers = json_decode( (string) file_get_contents( $root . '/includes/data/known-providers.json' ), true );
+
+el_eq( is_array( $providers ) && count( $providers ) > 100, true, 'the provider database loads' );
```
fabiodalez-dev: **suggestion:** The PR description says this guard "asserts the database is still 300+ entries", but the floor here is `100`. Given the database currently holds 300+ providers, the lower bound leaves a lot of room for the file to be gutted while the three Elementor checks still pass vacuously — which is the exact failure mode this assertion exists to prevent.

Tightening it to match the stated intent:

```suggestion
el_eq( is_array( $providers ) && count( $providers ) > 300, true, 'the provider database loads' );
```

tests/unit/test-elementor-not-blocked-php.php
```diff
@@ -0,0 +1,105 @@
...
+//    substring against the script URL. A site running Elementor serves the file
+//    from this path, so if any pattern is a substring of it the lightbox breaks
+//    again exactly as reported.
+$real_url  = '/wp-content/plugins/elementor/assets/lib/share-link/share-link.min.js';
+$would_block = array();
+foreach ( (array) $providers as $service_id => $service ) {
+	foreach ( (array) ( isset( $service['patterns'] ) ? $service['patterns'] : array() ) as $pattern ) {
+		$pattern = (string) $pattern;
+		if ( '' !== $pattern && false !== stripos( $real_url, $pattern ) ) {
+			$would_block[] = $service_id . ' → ' . $pattern;
```
fabiodalez-dev, line 83: **question:** Nice touch matching by substring the way the blocker does rather than by exact string — that's what makes this catch a re-introduction under a different spelling.

One consequence worth being deliberate about: this also fails if some *unrelated* provider ships a short, generic pattern that happens to be a substring of `/wp-content/plugins/elementor/assets/lib/share-link/share-link.min.js` (something like `/lib/` or `share`). I think that's correct behaviour — such a pattern really would break the lightbox again — but the failure message would point at the innocent-looking provider rather than at Elementor, and the next person to hit it may not immediately see why.

Could the label carry that hint? Something like `'the reported lightbox script URL matches no blocking pattern (a hit here means that provider's pattern is too broad)'` would save the reader a few minutes.

fabiodalez-dev (commented):
### Verdict
**question**

### Primary Goal
Stop the consent blocker from breaking Elementor's lightbox by removing the first-party `elementor/assets/lib/share-link` pattern from the shipped provider database and the Elementor blocker template, with a regression test pinning it.

### Overview
### What this PR does

Three changes, all pointed at issue #218:

1. **`includes/data/known-providers.json`** — drops the `elementor` provider entry (category `functional`, one pattern: `elementor/assets/lib/share-link`, zero cookies).
2. **`admin/modules/cookies/includes/blocker-templates/elementor.json`** — deletes the matching blocker template, whose description promised to block "Elementor embeds, widgets, or third-party content" while its only pattern was that same first-party helper.
3. **`tests/unit/test-elementor-not-blocked-php.php`** — a new standalone script that pins the removal three ways: no provider pattern names an Elementor asset path, the reported script URL matches no pattern **using the blocker's own substring semantics**, and no blocker template names one either.

### Assessment of the reasoning

The justification holds up. The pattern targets a file served from the site's own `plugins/` directory, the entry declared zero cookies, and `share-link` only builds share URLs — no storage, no third-party transmission, so there is nothing for consent to govern. Blocking it produced no privacy benefit and broke a core page-builder feature for every visitor who declined. Removal rather than re-categorisation is the right call here, and the decision *not* to generalise into a "zero cookies ⇒ never block" rule is well argued: entries like Site Kit and HubSpot share that shape but do go on to load real third parties.

The test is a good shape for this class of regression — matching by substring the way the blocker does, rather than by exact string, is what makes it catch a re-introduction under a different spelling. Asserting the database is still large also stops the checks from passing vacuously on an emptied file.

### Concerns

The code is fine. What is unresolved is **what happens to sites that already have this shipped rule persisted**:

- If `known-providers.json` is only read at runtime, this PR fixes every install on update and there is nothing more to do.
- If the provider list is seeded into an option (or a scanned-cookies table) on activation or on scan, then existing sites keep the stale pattern and stay broken — exactly the reporter's situation — until an upgrade routine removes it. The PR description covers admin-authored *custom* rules (correctly left alone) but is silent on shipped-defaults-copied-to-DB.

Second, smaller: a site that enabled the **Elementor blocker template** may hold `elementor` in an enabled-templates option. Worth confirming the template loader tolerates a stored id with no backing file rather than warning or fataling.

### Concerns list

- [ ] Confirm whether shipped provider entries are persisted to the DB; add an upgrade routine if so
- [ ] Confirm a stored `elementor` template id degrades gracefully now the file is gone
- [ ] Align the count floor in the test with the 300+ claimed in the description

### Scope Assessment
Tight and well-scoped. Two deletions that directly implement the fix, one test that guards it. No drift: no reformatting, no unrelated provider entries touched, no new dependencies, no changes to shared config or the blocking engine itself. The deliberate refusal to generalise the fix into a broad "zero cookies" rule is the right instinct and keeps the blast radius at exactly one provider.

### Risk Assessment
The privacy risk of removing a blocking rule is the direction that deserves scrutiny, and here it is genuinely near zero: first-party asset, no cookies, no background network call. The residual risk is **incompleteness rather than harm** — if the shipped provider database is seeded into the DB at activation or scan time, existing broken sites are not fixed by this change alone and need an upgrade routine. A secondary risk is a dangling `elementor` reference in a stored enabled-templates option now that the template file is gone.

### Reuse Notes
- The test deliberately reimplements the blocker's substring match (`stripos( $real_url, $pattern )`) rather than calling the production matcher. That is defensible for a standalone script with no WP bootstrap, but if a matching helper is reachable without loading WordPress, calling it would keep the test honest if the matching semantics ever change (e.g. to regex or word-boundary matching).
- The third parties Elementor can actually embed — YouTube, Vimeo, Google Maps, reCAPTCHA — already have their own provider entries, so no coverage is lost by dropping the Elementor template. Worth stating in the changelog so admins who had the template enabled know they are not newly exposed.

### Action Items
- [ ] Answer the migration question: is `known-providers.json` read live, or copied into an option/table on activation or scan? If copied, add an upgrade routine that strips the `elementor` entry from shipped (not custom) rules.
- [ ] Verify a stored `elementor` blocker-template id no longer backed by a file does not warn or fatal in the templates admin screen.
- [ ] Raise the provider-count floor in the test from `> 100` to `> 300` to match the guard described in the PR body.
- [ ] Confirm the new test file exits non-zero when a check fails, so CI actually fails on a regression rather than printing red text and passing.
- [ ] Add a changelog line noting that Elementor's own assets are no longer blocked, and that real third-party embeds remain covered by their own provider entries.

### Separate PR Suggestions
- If the shipped provider database turns out to be persisted to the DB, the upgrade routine that strips the stale `elementor` entry from existing installs is worth its own PR — it touches migration code and deserves its own testing and review, separate from this data-only removal.

PR #228, opened 2026-08-13
Refactor _fazRestoreInlineScript to eliminate eval() and fix CSP violations

rwkyyy:
**Problem**
`_fazRestoreInlineScript()` currently relies on `(0, eval)(inlineText)` to execute inline scripts once consent is granted. Mutating the `type` attribute on an existing `<script>` element after parsing does not trigger browser execution, which led to using `eval()` as a workaround. This causes high volumes of CSP violations on environments enforcing `script-src` policies without `'unsafe-eval'`.

**Solution**

* Replaced the indirect `eval()` call in `_fazRestoreInlineScript()` by delegating to the existing `_fazBuildRestoredScript()` helper.
* Instead of executing code in-place via `eval()`, the function now constructs a fresh `<script>` clone and replaces the original node in the DOM using `script.parentNode.replaceChild(clone, script)`.
* Inserting a new `<script>` element into the DOM triggers native browser parsing/execution safely without requiring `eval()`.
* Source (`frontend/js/script.js`) updated and minified build (`frontend/js/script.min.js`) rebuilt via Terser.

**Testing**

* Confirmed blocked inline scripts (GTM, Meta Pixel, etc.) execute correctly upon accepting cookies.
* Verified in Browser DevTools that CSP `[Report Only]` violations for `eval` no longer fire on consent restoration.

## Summary by CodeRabbit

* **Miglioramenti**
  * Aggiornato il ripristino degli script inline senza ricorrere a `eval`.
  * Preservato il nonce CSP durante la riattivazione degli script eseguibili.
  * I contenuti JSON e gli script non eseguibili vengono ripristinati direttamente senza esecuzione di codice.
  * Evitata l’esecuzione ripetuta degli script già ripristinati e gestiti correttamente anche gli elementi non più presenti nella pagina.


fabiodalez-dev:
Thanks — this is the right fix. Indirect `eval()` is refused outright by any CSP without `'unsafe-eval'`, so on exactly the strictest sites consented scripts were never running. Rebuilding the node and letting the browser execute it on insertion is the correct mechanism.

I've pushed three things on top of your commit:

**The CSP nonce.** Once an element is connected to the document, the browser clears the `nonce` content attribute and keeps the value only on the IDL property — deliberately, so a CSS attribute selector can't read it back out. The attribute-copy loop in `_fazBuildRestoredScript` therefore saw an empty string, and the clone arrived without a nonce: blocked by the very policy this change exists to survive. Now copied through `script.nonce`.

**JSON payloads still need unblocking.** They're never executed, but they do need their type back. The old code restored the type *before* it looked at the payload, so that branch inherited the restoration for free; delegating to `_fazBuildRestoredScript` moved it out of reach, and an `application/ld+json` block stayed at the placeholder type after consent — structured data invisible to anything reading the DOM.

**The order inside that restoration**, which turned out to be a pre-existing bug your change surfaced rather than one it introduced. The blocker intercepts type changes, so a script that turns runnable while still carrying `data-faz-category` is precisely what it's built to stop: it re-applies the placeholder and the restore silently undoes itself. Markers come off first, then the type goes back.

I added `tests/unit/js/inline-script-restore.test.mjs` — 13 checks, run against three trees to be sure they aren't passing by construction: 3 fail against your branch before the fix, 5 against `main`, 0 now. The fixtures sit in a detached parent on purpose; attached ones get re-blocked by the live observer, and would have measured the blocker instead of the restore path.

Going into 1.26.0.

PR #239, opened 2026-08-19
fix(scanner,amp): make a failed import retryable, and align AMP purpose ids

fabiodalez-dev:
Closes the last three findings from the full review of `codex/fix-scanner-stale-safety`, plus the 1.27.0 release metadata. Stacked on that branch, so the diff here is only the new work.

## A failed scan import destroyed the evidence it told you to retry with

`import_cookies()` performed two irreversible side effects before the try/save that can fail:

- `finish_browser_scan_session()` both **drained** the observations and **tore down** the session — deleting every user-meta observation, both transients and the marker cookie — ninety lines before `save_scan_result()` ran.
- `schedule_httponly_check()` queued a full background replay of every scanned URL before the same try.

So a save failure destroyed the run's server-captured HttpOnly observations, the retry the 500 advertised answered 409 because the session was gone, and a cron replay still wrote catalogue rows for an import that never completed.

The split is now: `collect_browser_scan_session()` drains without deleting, `finish_browser_scan_session()` is the single teardown, `abort_browser_scan_session()` delegates to it so there is exactly one deletion site per resource, and both the teardown and the replay happen only after `save_scan_result()` returns. The catch does neither, which is what makes its "try again" true.

Verified with a harness that subclasses the **real** `Controller` — real transients, real user meta, real marker — and throws once from `save_scan_result()`: the session survives the failure, no replay is scheduled, the second import with the same `scan_id` succeeds and persists the server-captured HttpOnly row, and only then is the session closed and one replay queued.

## …and the browser then undid it one request later

`scan-engine.js` handled a failed import with `rejectAndAbort()`, which POSTs `scans/abort`, which is now literally the teardown the server had just deferred. The observations were destroyed milliseconds after the 500 — so the fix was correct in PHP and absent from the product.

An import-stage failure no longer aborts. Every other stage still does, since holding the capture lock open after a network or iframe failure has no upside.

## AMP purpose ids spoke two vocabularies

`purposeConsentRequired` and `data-block-on-consent-purposes` emitted the raw category slug while the consent checkbox called `setPurpose()` with the underscore-translated action-argument name. For a hyphenated slug the visitor's tick was recorded under one id and the gate tested another, so a gated component could never unblock after "Save preferences" — over-blocking, on the one surface where the visitor had just said yes.

All three surfaces plus the REST responses now use `purpose_action_args()` output. `normalize_purpose_consent()` still returns slug-keyed data, reading the arg name first with the slug as fallback, so cookies and signed state strings written before this change still decode — no migration. The `faz_amp_component_purpose_map` filter keeps receiving slugs, preserving its published contract.

## Four assertions that could not fail

Found by verifying the above rather than trusting it, and called out because it is a recurring shape here:

- `FazTest_Import_Controller` stubbed `browser_scan_session_matches()` to `true` unconditionally, so *"the same scan id can retry successfully after persistence recovers"* stayed green with the fix **fully reverted** — the fake was structurally unable to express "the session is gone", which is the entire failure mode. It now models the session's lifetime; reintroducing the regression turns five checks red.
- The teardown-ordering check compared `strpos` offsets, i.e. the *first* occurrence, so a second teardown added inside the `catch` — the exact regression it is named after — passed. The catch block is now brace-matched and asserted inert.
- `4 === count( $safe_urls )` read as a check on URL normalisation and was not one; it went 2 → 4 because two URLs were added to the *input*. Replaced with the property the filter exists for: four same-site URLs kept, four hostile shapes refused, each named.

## Also here

- `normalize_url()` no longer appends a trailing slash to a file-like path, so a site on a `/%postname%.html` permalink structure is scanned at its real URLs instead of 404s.
- A negative `Max-Age` is read as an expiry rather than `absint()`'d into a positive lifetime.
- 1.27.0 version and changelog metadata.

## Verification

110/110 unit suites. E2E: `release-verify-amp-bridge` 6/6, `release-verify-scan-session` 6/6, `release-verify-admin-misc` 8/8.

One pre-existing failure is **not** from this branch: `scan-catalog-deep` test 05 times out, and it fails identically with the pre-change PHP files deployed.


fabiodalez-dev:
### Code review

**Branch:** `fix/scanner-retry-amp-purposes` → `codex/fix-scanner-stale-safety`
**PR:** #239 (open)
**Review ID:** `rev_20260819T171716Zed2d7a`
**Sub-agent tokens:** 0 across 0 invocations

Found 18 findings across all lanes:
- Deep lane (correctness/security): 6 resolved

---

### ✓ Auto-fixable (6)

| # | Score | Impact | File | Issue | Status |
|---|-------|--------|------|-------|--------|
| F001 | 58 | correctness | `admin/assets/js/modules/scan-engine.js:612-617` | The PR's headline goal — a failed import leaves the capture session intact so the evidence can be retried — is defeated by the client. After importWithRetry exhausts its two automatic attempts the catch calls rejectAndAbort(), which POSTs scans/abort, which delegates to finish_browser_scan_session() and deletes every persisted observation and the marker cookie. The server-side preservation is undone one round-trip later on every terminal failure. tests/unit/js/scan-engine.test.mjs check 30 asserts the abort HAPPENS, so the test suite pins the behaviour the PR set out to remove. | ✓ fixed and verified (`d2396cd`) |
| F003 | 62 | correctness | `admin/modules/scanner/includes/class-controller.php:301-313` | set_cookie_is_deletion() is a point-in-time test — for the Expires branch it compares strtotime($expires) <= time() — but both callers apply it to ALREADY PERSISTED rows on every later scan-tagged request. A cookie legitimately set with an absolute Expires inside the crawl window (a CAPTCHA challenge token, a rate-limit counter, a flash cookie) is re-read minutes later, reclassified as a clearing directive, and delete_user_meta()'d permanently. The row carries its own observed_at (:358) and the comparison should use it. A cookie the site really sets pre-consent silently disappears from the public declaration — the exact under-reporting direction this scanner exists to prevent. | ✓ fixed and verified (`a825a90`) |
| F004 | 52 | correctness | `admin/modules/scanner/includes/class-controller.php:746-757` | Sibling validators disagree on a malformed Max-Age. set_cookie_is_deletion() deliberately gates it on ^-?[0-9]+$ and lets a non-numeric value fall through to Expires, but the duration computation beside it enters the max-age branch on any non-empty value, casts 'abc' to 0, fails the >0 test and never evaluates the elseif — so 'Max-Age=abc; Expires=<two years>' is declared 'session'. The server replay path at :1806 has the same bug AND inverts the pre-PR precedence (Expires was checked first at old :1730), making it a regression, not just an inconsistency. | ✓ fixed and verified (`a825a90`) |
| F005 | 55 | correctness | `admin/assets/js/modules/scan-engine.js:568-585` | importIsRetryable() treats HTTP status 0 as retryable, but status 0 (network drop, navigation, timeout with no response) is precisely the case where the request may have REACHED the server and succeeded. Attempt 1 then already ran the teardown, so attempt 2 gets a 409 expired session and the run rejects for a scan that actually imported. The failure mode did not exist before the retry was added. | ✓ fixed and verified (`d2396cd`) |
| F006 | 58 | correctness | `admin/assets/js/modules/scan-engine.js:145-152` | The client's normalizeUrl() was not updated to match the new PHP normalize_url() rule, so the two now disagree. discover returns '/wp-sitemap.xml' unslashed, the client re-appends the slash, crawls and reports the slashed variant, and the replay queues the slashed variant — the paths this PR stopped rewriting server-side are still rewritten before anything fetches them, and the capture keys diverge between the two halves. | ✓ fixed and verified (`a825a90`) |
| F007 | 60 | correctness | `admin/modules/scanner/includes/class-controller.php:316-320` | home_url( $request_path ) double-prefixes the subdirectory on a WordPress installed in one: REQUEST_URI's path already contains '/blog', so '/blog/shop/x' becomes 'https://site/blog/blog/shop/x'. set_cookie_scope() then derives a wrong RFC 6265 default-path, which is persisted and baked into set_cookie_identity(). Both the writer (:305) and the reader (:723) are wrong in the same direction, which is what hides it — but an explicitly-pathed cookie compared against a default-pathed one no longer matches, so an ordered clearing directive can fail to cancel its prior sighting. | ✓ fixed and verified (`a825a90`) |

**Cross-cutting group G1:** F003 + F004 + F007 — All three live in the cookie-identity/lifetime helpers this PR introduced, and all three are the same mistake: the code evaluates against the wrong reference. F003 compares an expiry against the current time when the row carries observed_at; F004 applies a looser notion of a valid Max-Age than the sibling validator added in the same commit; F007 re-prefixes a home path that the request path already contains. Fix them as one pass over that new code, because they share a review question - which reference does this line mean? - and a single test fixture (persist one observation, advance the clock, re-read) exercises all three. Treating them separately invites a fix for F003 that keeps reading the wrong URL, or a Max-Age guard added to one of two sibling branches again. Order: F007 first (it changes the persisted key, so it must land before any test pins that key), then F003, then F004.
**Cross-cutting group G2:** F001 + F005 — Both concern what happens after an import fails, and fixing either alone makes the other worse. F001: the client aborts once the two automatic retries are exhausted, destroying the session the server deliberately kept - contradicting the commit message. F005: a status-0 failure after a server-side success guarantees a 409 on the next attempt, because the session was already finished. Holding the session (F001) without an idempotent resubmit (F005) preserves data an administrator still cannot import; making the resubmit idempotent without holding the session preserves nothing to resubmit against. Decide the pair together: either implement both, or implement neither and correct the commit message plus the three comments that promise a hold. The honest minimum is the second - it costs nothing and stops the code from claiming a guarantee it does not offer.

#### F001 — The PR's headline goal — a failed import leaves the capture session intact so the evidence can be retried — is defeated by the client. After importWithRetry exhausts its two automatic attempts the catch calls rejectAndAbort(), which POSTs scans/abort, which delegates to finish_browser_scan_session() and deletes every persisted observation and the marker cookie. The server-side preservation is undone one round-trip later on every terminal failure. tests/unit/js/scan-engine.test.mjs check 30 asserts the abort HAPPENS, so the test suite pins the behaviour the PR set out to remove.

**File:** `admin/assets/js/modules/scan-engine.js:612-617`
**Score:** 58 (moderate)
**Reason:** CONFIRMED: Verified: scan-engine.js:612-617 sets stage='import' then calls rejectAndAbort, which POSTs scans/abort (:337). api/class-api.php:921-923 -> abort_browser_scan_session (controller:838-840) -> finish_browser_scan_session (:780-808), deleting observations, transients and marker. DECISIVE: commit ee81ad0's own message states 'An import-stage failure no longer aborts ... worth holding until the idle TTL expires' - the code and test check 30 do the opposite. || Scope correction: Not fully defeated: the two automatic retries DO consume the preserved session (check 28, aborts.length===0). Only the post-exhaustion hold is missing. Code and tests are SELF-CONSISTENT; the mismatch is with the commit message.

**Evidence:**
- Verified: scan-engine.js:612-617 sets stage='import' then calls rejectAndAbort, which POSTs scans/abort (:337). api/class-api.php:921-923 -> abort_browser_scan_session (controller:838-840) -> finish_browser_scan_session (:780-808), deleting observations, transients and marker. DECISIVE: commit ee81ad0's own message states 'An import-stage failure no longer aborts ... worth holding until the idle TTL expires' - the code and test check 30 do the opposite.
- Not fully defeated: the two automatic retries DO consume the preserved session (check 28, aborts.length===0). Only the post-exhaustion hold is missing. Code and tests are SELF-CONSISTENT; the mismatch is with the commit message.

**Approach:** Decide between the commit message and the code: either hold the session after the retries are exhausted, or stop claiming that it is held.

**Files to modify:**
- `admin/assets/js/modules/scan-engine.js` — Either drop the abort from the import-stage catch at :612-617 and let BROWSER_SCAN_TTL reclaim the session, or keep the abort and correct the commit message plus the three comments that describe a hold. (why: ee81ad0's message states an import-stage failure 'no longer aborts ... worth holding until the idle TTL expires'; the code aborts. Code and check 30 agree with each other, so only the description is wrong - but holding the data is only worth doing if an administrator can reach it.)

**Verification:**
- If holding: assert no scans/abort is POSTed after the retries are exhausted, and that the observations survive.
- If not holding: assert the commit message and comments no longer promise a hold.

**Edge cases to preserve:**
- The two automatic retries must keep working against the intact session (check 28).

**Latest fix attempt (fixrun_20260820T011724Z3e2e4e):** fixed and verified
- Phase 9 finding: Resolved by explicit user decision to implement the hold rather than correct the commit message. The import-stage catch now holds instead of aborting (discover/iframe still abort). Crucially the hold alone would have LOCKED THE ADMINISTRATOR OUT for the capture TTL - which is exactly what rejectAndAbort's comment says it existed to prevent - so a session is now marked state='held' and start_browser_scan_session() reclaims a held session while still 409-ing a live one. scan-engine.test.mjs check 30, which asserted the abort, was deliberately inverted; check 28 untouched.

#### F003 — set_cookie_is_deletion() is a point-in-time test — for the Expires branch it compares strtotime($expires) <= time() — but both callers apply it to ALREADY PERSISTED rows on every later scan-tagged request. A cookie legitimately set with an absolute Expires inside the crawl window (a CAPTCHA challenge token, a rate-limit counter, a flash cookie) is re-read minutes later, reclassified as a clearing directive, and delete_user_meta()'d permanently. The row carries its own observed_at (:358) and the comparison should use it. A cookie the site really sets pre-consent silently disappears from the public declaration — the exact under-reporting direction this scanner exists to prevent.

**File:** `admin/modules/scanner/includes/class-controller.php:301-313`
**Score:** 62 (moderate)
**Reason:** CONFIRMED: Verified. set_cookie_is_deletion (:1905-1915) compares strtotime($expires) <= time(); rows persist the RAW header (:364) and are re-evaluated on every later scan-tagged request. Writer :303-313 calls delete_user_meta on the row itself (:310) plus prior same-key rows (:305); reader :725 silently unsets it. Window is the session (sliding 900s, ceiling BROWSER_SCAN_MAX_AGE=21600 at :135), not 15 minutes. || Scope correction: Observer reads headers_list() - this site's own PHP output. setcookie() always emits Max-Age beside expires, and Max-Age short-circuits at :1907. Only raw header('Set-Cookie: ...Expires') cookies are exposed, so the CAPTCHA/flash examples are mostly immune.
**Auto-recommendation (medium):** Compare the cookie's expiry against the observation's own timestamp instead of the current time.

**Evidence:**
- Verified. set_cookie_is_deletion (:1905-1915) compares strtotime($expires) <= time(); rows persist the RAW header (:364) and are re-evaluated on every later scan-tagged request. Writer :303-313 calls delete_user_meta on the row itself (:310) plus prior same-key rows (:305); reader :725 silently unsets it. Window is the session (sliding 900s, ceiling BROWSER_SCAN_MAX_AGE=21600 at :135), not 15 minutes.
- Observer reads headers_list() - this site's own PHP output. setcookie() always emits Max-Age beside expires, and Max-Age short-circuits at :1907. Only raw header('Set-Cookie: ...Expires') cookies are exposed, so the CAPTCHA/flash examples are mostly immune.

**Approach:** Compare the cookie's expiry against the observation's own timestamp instead of the current time.

**Files to modify:**
- `admin/modules/scanner/includes/class-controller.php` — set_cookie_is_deletion() takes the reference time as a parameter; both callers pass the row's observed_at (present at :358) and the live path passes time(). (why: A predicate that is correct on a freshly received header is wrong on a row persisted minutes ago; the record already carries the timestamp the comparison needs.)

**Verification:**
- Persist a row with Expires = now+120s, advance past it, re-run a scan-tagged request, assert the row survives.
- Assert a genuine clearing directive (Expires in the past AT OBSERVATION TIME) is still removed.

**Edge cases to preserve:**
- Max-Age<=0 must still mean deletion regardless of timing.
- A directive observed as already-expired must still delete.

**Latest fix attempt (fixrun_20260820T011724Z3e2e4e):** fixed and verified
- Phase 9 finding: Resolved. set_cookie_is_deletion() gained an optional $reference_time; persisted-row callers (:303, :725) pass observation_reference_time(), live-header callers (:331, :1740, :1785) keep time(). Phase 9a audited all five callsites. The duration computation at :757 was ALSO moved onto the same reference after 9a flagged it - fixing the deletion check alone would have let the surviving cookie be reported as 'session'.

#### F004 — Sibling validators disagree on a malformed Max-Age. set_cookie_is_deletion() deliberately gates it on ^-?[0-9]+$ and lets a non-numeric value fall through to Expires, but the duration computation beside it enters the max-age branch on any non-empty value, casts 'abc' to 0, fails the >0 test and never evaluates the elseif — so 'Max-Age=abc; Expires=<two years>' is declared 'session'. The server replay path at :1806 has the same bug AND inverts the pre-PR precedence (Expires was checked first at old :1730), making it a regression, not just an inconsistency.

**File:** `admin/modules/scanner/includes/class-controller.php:746-757`
**Score:** 52 (moderate)
**Reason:** CONFIRMED: Verified. :1808 enters the max-age branch on any non-empty value; (int)'abc'=0 fails >0 and the elseif at :1813 never runs. No upstream sanitisation - parse_set_cookie (:1959) stores raw attribute text, and scan_page headers are remote-controlled. || Scope correction: The precedence inversion is RFC 6265 5.3-CORRECT and matches the deliberate design at :1899. Base was equally wrong (absint('abc')=0), just reached Expires first - so NOT a regression. Impact is a duration label; :747's value is PHP-generated, so only :1808 is realistically reachable.
**Auto-recommendation (medium):** Mirror set_cookie_is_deletion()'s Max-Age validator in the duration computation so a malformed value defers to Expires.

**Evidence:**
- Verified. :1808 enters the max-age branch on any non-empty value; (int)'abc'=0 fails >0 and the elseif at :1813 never runs. No upstream sanitisation - parse_set_cookie (:1959) stores raw attribute text, and scan_page headers are remote-controlled.
- The precedence inversion is RFC 6265 5.3-CORRECT and matches the deliberate design at :1899. Base was equally wrong (absint('abc')=0), just reached Expires first - so NOT a regression. Impact is a duration label; :747's value is PHP-generated, so only :1808 is realistically reachable.

**Approach:** Mirror set_cookie_is_deletion()'s Max-Age validator in the duration computation so a malformed value defers to Expires.

**Files to modify:**
- `admin/modules/scanner/includes/class-controller.php` — Gate the max-age branch at :1808 (and :747) on the same ^-?[0-9]+$ regex used at :1907, so a non-numeric value falls through to the Expires branch. (why: Two sibling helpers currently disagree about what a valid Max-Age is; the PR added the validator to one of them only.)

**Verification:**
- Feed 'Max-Age=abc; Expires=<two years>' through scan_page() and assert the duration is two years, not 'session'.

**Edge cases to preserve:**
- A valid Max-Age must keep precedence over Expires - that ordering is RFC 6265 5.3-correct and deliberate.

**Latest fix attempt (fixrun_20260820T011724Z3e2e4e):** fixed and verified
- Phase 9 finding: Resolved. Both sites (:749, :1814) now trim and gate Max-Age on the same ^-?[0-9]+$ used by set_cookie_is_deletion(). 9a verified in situ incl. the negative-Max-Age path. No comment calls it a regression.

#### F005 — importIsRetryable() treats HTTP status 0 as retryable, but status 0 (network drop, navigation, timeout with no response) is precisely the case where the request may have REACHED the server and succeeded. Attempt 1 then already ran the teardown, so attempt 2 gets a 409 expired session and the run rejects for a scan that actually imported. The failure mode did not exist before the retry was added.

**File:** `admin/assets/js/modules/scan-engine.js:568-585`
**Score:** 55 (moderate)
**Reason:** CONFIRMED: Verified: import_cookies has NO idempotency. A second import for the same scan_id fails browser_scan_session_matches() and returns 409 faz_browser_scan_session_expired (class-api.php:713-731), because finish_browser_scan_session() runs after a successful save (:840). getApiErrorStatus() returns 0 for any error lacking a numeric status (scan-engine.js:167-172), so a lost response after a server-side SUCCESS is retried and the run rejects (:611-616) despite the data being imported. || Scope correction: NOT a regression: without retry-on-0 that scenario failed identically. Excluding status 0 would forfeit real recovery - the fix belongs in import IDEMPOTENCY (F011), not the retry predicate.

**Evidence:**
- Verified: import_cookies has NO idempotency. A second import for the same scan_id fails browser_scan_session_matches() and returns 409 faz_browser_scan_session_expired (class-api.php:713-731), because finish_browser_scan_session() runs after a successful save (:840). getApiErrorStatus() returns 0 for any error lacking a numeric status (scan-engine.js:167-172), so a lost response after a server-side SUCCESS is retried and the run rejects (:611-616) despite the data being imported.
- NOT a regression: without retry-on-0 that scenario failed identically. Excluding status 0 would forfeit real recovery - the fix belongs in import IDEMPOTENCY (F011), not the retry predicate.

**Approach:** Make the import idempotent for a repeated scan_id rather than removing status 0 from the retry predicate.

**Files to modify:**
- `admin/modules/scanner/api/class-api.php` — Let a resubmit for a scan_id whose session was just finished return the previous success instead of 409 faz_browser_scan_session_expired (:713-731). (why: A status-0 failure is exactly the case where the request reached the server and succeeded; attempt 1 already ran finish_browser_scan_session() at :840, so attempt 2 is guaranteed to 409 and the run reports failure for a scan that imported.)

**Verification:**
- Import successfully, then resubmit the identical payload, and assert the response is a success, not a 409.

**Edge cases to preserve:**
- A genuinely expired session must still be rejected.
- Excluding status 0 from the retry predicate would forfeit real recovery - do not take that route.

**Latest fix attempt (fixrun_20260820T011724Z3e2e4e):** fixed and verified
- Phase 9 finding: Resolved. Rather than narrowing the retry predicate (the validator's advice: excluding status 0 would forfeit real recovery), the import is now idempotent. The outcome is remembered against the scan id and a resubmit is answered with it, marked duplicate, with no second save and no second background replay. Write order is load-bearing and pinned by a mutation-proved test: remember BEFORE finish, so a request dying between them still answers a resubmit truthfully.

#### F006 — The client's normalizeUrl() was not updated to match the new PHP normalize_url() rule, so the two now disagree. discover returns '/wp-sitemap.xml' unslashed, the client re-appends the slash, crawls and reports the slashed variant, and the replay queues the slashed variant — the paths this PR stopped rewriting server-side are still rewritten before anything fetches them, and the capture keys diverge between the two halves.

**File:** `admin/assets/js/modules/scan-engine.js:145-152`
**Score:** 58 (moderate)
**Reason:** CONFIRMED: Client normalizeUrl (:145-151) still forces a trailing slash and IS applied to server-supplied URLs via deduplicateUrls(result.urls / result.priority_urls) at :411-414; the slashed form is crawled, pushed to metrics.scannedUrls (:753), sent as scanned_urls (:563) and re-queued by sanitize_scanned_urls (class-controller.php:1170). PHP's new heuristic (:1407-1425) leaves file-like paths unslashed, so the divergence is real and the fix is undone on the browser path. || Scope correction: My /wp-sitemap.xml example was WRONG: discover_urls uses discover_pages_from_db/get_priority_urls (permalinks only), never the sitemap. Real trigger is a file-extension permalink structure such as /%postname%.html.
**Auto-recommendation (high):** Align the client's normalizeUrl() with the PHP rule the PR introduced.

**Evidence:**
- Client normalizeUrl (:145-151) still forces a trailing slash and IS applied to server-supplied URLs via deduplicateUrls(result.urls / result.priority_urls) at :411-414; the slashed form is crawled, pushed to metrics.scannedUrls (:753), sent as scanned_urls (:563) and re-queued by sanitize_scanned_urls (class-controller.php:1170). PHP's new heuristic (:1407-1425) leaves file-like paths unslashed, so the divergence is real and the fix is undone on the browser path.
- My /wp-sitemap.xml example was WRONG: discover_urls uses discover_pages_from_db/get_priority_urls (permalinks only), never the sitemap. Real trigger is a file-extension permalink structure such as /%postname%.html.

**Approach:** Align the client's normalizeUrl() with the PHP rule the PR introduced.

**Files to modify:**
- `admin/assets/js/modules/scan-engine.js` — Apply the same file-like-path exemption at :145-152 that class-controller.php:1407-1425 now uses, so the client stops re-appending a trailing slash to paths the server deliberately left bare. (why: deduplicateUrls() applies the client rule to server-supplied URLs at :411-414, so the browser path undoes the server-side fix before anything is fetched.)

**Verification:**
- On a /%postname%.html permalink structure, assert the URL crawled and reported matches the URL discover returned, byte for byte.

**Edge cases to preserve:**
- Directory paths must still be slashed; /.well-known/ must stay a directory.

**Latest fix attempt (fixrun_20260820T011724Z3e2e4e):** fixed and verified
- Phase 9 finding: Resolved. Client normalizeUrl() mirrors the PHP rule incl. the leading-dot exclusion. Equal on all 13 fixture cases in both runners; 9a mutation-tested both.

#### F007 — home_url( $request_path ) double-prefixes the subdirectory on a WordPress installed in one: REQUEST_URI's path already contains '/blog', so '/blog/shop/x' becomes 'https://site/blog/blog/shop/x'. set_cookie_scope() then derives a wrong RFC 6265 default-path, which is persisted and baked into set_cookie_identity(). Both the writer (:305) and the reader (:723) are wrong in the same direction, which is what hides it — but an explicitly-pathed cookie compared against a default-pathed one no longer matches, so an ordered clearing directive can fail to cancel its prior sighting.

**File:** `admin/modules/scanner/includes/class-controller.php:316-320`
**Score:** 60 (moderate)
**Reason:** CONFIRMED: Confirmed. wp-includes/link-template.php:3481-3483 appends the path to get_option('home') VERBATIM, so on a subdirectory install home_url('/blog/shop/x') yields '.../blog/blog/shop/x'. class-controller.php:316-320 and :722-724 feed exactly REQUEST_URI's path, and the doubled URL reaches set_cookie_scope()'s RFC 6265 default-path branch (:1867-1875), persisted at :360-361 and keyed by set_cookie_identity(). Both callsites are NEW in this PR. || Scope correction: Host unaffected; only the default-path branch (no explicit Path) is wrong, and only on subdirectory installs. Read/write use the same wrong URL, so damage is mismatching explicit-Path vs default-Path rows plus the stored path.
**Auto-recommendation (high):** Build the observation URL from the site root plus the request path, without re-prefixing the home path.

**Evidence:**
- Confirmed. wp-includes/link-template.php:3481-3483 appends the path to get_option('home') VERBATIM, so on a subdirectory install home_url('/blog/shop/x') yields '.../blog/blog/shop/x'. class-controller.php:316-320 and :722-724 feed exactly REQUEST_URI's path, and the doubled URL reaches set_cookie_scope()'s RFC 6265 default-path branch (:1867-1875), persisted at :360-361 and keyed by set_cookie_identity(). Both callsites are NEW in this PR.
- Host unaffected; only the default-path branch (no explicit Path) is wrong, and only on subdirectory installs. Read/write use the same wrong URL, so damage is mismatching explicit-Path vs default-Path rows plus the stored path.

**Approach:** Build the observation URL from the site root plus the request path, without re-prefixing the home path.

**Files to modify:**
- `admin/modules/scanner/includes/class-controller.php` — At :316-320 and :722-724 stop passing REQUEST_URI's full path to home_url(); that path already contains the subdirectory, which home_url() prepends again. (why: WP core link-template.php:3481-3483 appends the argument to get_option('home') verbatim, so a subdirectory install gets /blog/blog/... and set_cookie_scope() derives a wrong RFC 6265 default-path that is persisted and keyed.)

**Verification:**
- On a subdirectory install, assert the stored path for a default-path cookie equals the real request directory.

**Edge cases to preserve:**
- Root installs must stay byte-identical.
- The host component is already correct - do not change it.

**Latest fix attempt (fixrun_20260820T011724Z3e2e4e):** fixed and verified
- Phase 9 finding: Resolved. observation_url() builds from the site root. THREE callsites converted (:301, :321, :723) - the third was not in the finding but keys the same persisted rows. Root installs byte-identical, host and port preserved.

</details>

### Auto-recommendations (4)

_AI-authored fix directions for findings that aren't auto-fixable today. Run `/adamsreview:fix` to batch-apply with one confirmation, or `/adamsreview:walkthrough` to review one-by-one._

| # | Score | Disp | File | Recommendation |
|---|-------|------|------|----------------|
| F003 | 62 | resolved | `admin/modules/scanner/includes/class-controller.php:301-313` | _medium:_ Compare the cookie's expiry against the observation's own timestamp instead of the current time. |
| F004 | 52 | resolved | `admin/modules/scanner/includes/class-controller.php:746-757` | _medium:_ Mirror set_cookie_is_deletion()'s Max-Age validator in the duration computation so a malformed value defers to Expires. |
| F006 | 58 | resolved | `admin/assets/js/modules/scan-engine.js:145-152` | _high:_ Align the client's normalizeUrl() with the PHP rule the PR introduced. |
| F007 | 60 | resolved | `admin/modules/scanner/includes/class-controller.php:316-320` | _high:_ Build the observation URL from the site root plus the request path, without re-prefixing the home path. |

<details><summary>Full recommendations and alternatives</summary>

#### F003 — set_cookie_is_deletion() is a point-in-time test — for the Expires branch it compares strtotime($expires) <= time() — but both callers apply it to ALREADY PERSISTED rows on every later scan-tagged request. A cookie legitimately set with an absolute Expires inside the crawl window (a CAPTCHA challenge token, a rate-limit counter, a flash cookie) is re-read minutes later, reclassified as a clearing directive, and delete_user_meta()'d permanently. The row carries its own observed_at (:358) and the comparison should use it. A cookie the site really sets pre-consent silently disappears from the public declaration — the exact under-reporting direction this scanner exists to prevent.

**File:** `admin/modules/scanner/includes/class-controller.php:301-313`
**Score:** 62 (moderate)
**Reason:** CONFIRMED: Verified. set_cookie_is_deletion (:1905-1915) compares strtotime($expires) <= time(); rows persist the RAW header (:364) and are re-evaluated on every later scan-tagged request. Writer :303-313 calls delete_user_meta on the row itself (:310) plus prior same-key rows (:305); reader :725 silently unsets it. Window is the session (sliding 900s, ceiling BROWSER_SCAN_MAX_AGE=21600 at :135), not 15 minutes. || Scope correction: Observer reads headers_list() - this site's own PHP output. setcookie() always emits Max-Age beside expires, and Max-Age short-circuits at :1907. Only raw header('Set-Cookie: ...Expires') cookies are exposed, so the CAPTCHA/flash examples are mostly immune.
**Auto-recommendation (medium):** Compare the cookie's expiry against the observation's own timestamp instead of the current time.

**Evidence:**
- Verified. set_cookie_is_deletion (:1905-1915) compares strtotime($expires) <= time(); rows persist the RAW header (:364) and are re-evaluated on every later scan-tagged request. Writer :303-313 calls delete_user_meta on the row itself (:310) plus prior same-key rows (:305); reader :725 silently unsets it. Window is the session (sliding 900s, ceiling BROWSER_SCAN_MAX_AGE=21600 at :135), not 15 minutes.
- Observer reads headers_list() - this site's own PHP output. setcookie() always emits Max-Age beside expires, and Max-Age short-circuits at :1907. Only raw header('Set-Cookie: ...Expires') cookies are exposed, so the CAPTCHA/flash examples are mostly immune.

**Approach:** Compare the cookie's expiry against the observation's own timestamp instead of the current time.

**Files to modify:**
- `admin/modules/scanner/includes/class-controller.php` — set_cookie_is_deletion() takes the reference time as a parameter; both callers pass the row's observed_at (present at :358) and the live path passes time(). (why: A predicate that is correct on a freshly received header is wrong on a row persisted minutes ago; the record already carries the timestamp the comparison needs.)

**Verification:**
- Persist a row with Expires = now+120s, advance past it, re-run a scan-tagged request, assert the row survives.
- Assert a genuine clearing directive (Expires in the past AT OBSERVATION TIME) is still removed.

**Edge cases to preserve:**
- Max-Age<=0 must still mean deletion regardless of timing.
- A directive observed as already-expired must still delete.

**Latest fix attempt (fixrun_20260820T011724Z3e2e4e):** fixed and verified
- Phase 9 finding: Resolved. set_cookie_is_deletion() gained an optional $reference_time; persisted-row callers (:303, :725) pass observation_reference_time(), live-header callers (:331, :1740, :1785) keep time(). Phase 9a audited all five callsites. The duration computation at :757 was ALSO moved onto the same reference after 9a flagged it - fixing the deletion check alone would have let the surviving cookie be reported as 'session'.

#### F004 — Sibling validators disagree on a malformed Max-Age. set_cookie_is_deletion() deliberately gates it on ^-?[0-9]+$ and lets a non-numeric value fall through to Expires, but the duration computation beside it enters the max-age branch on any non-empty value, casts 'abc' to 0, fails the >0 test and never evaluates the elseif — so 'Max-Age=abc; Expires=<two years>' is declared 'session'. The server replay path at :1806 has the same bug AND inverts the pre-PR precedence (Expires was checked first at old :1730), making it a regression, not just an inconsistency.

**File:** `admin/modules/scanner/includes/class-controller.php:746-757`
**Score:** 52 (moderate)
**Reason:** CONFIRMED: Verified. :1808 enters the max-age branch on any non-empty value; (int)'abc'=0 fails >0 and the elseif at :1813 never runs. No upstream sanitisation - parse_set_cookie (:1959) stores raw attribute text, and scan_page headers are remote-controlled. || Scope correction: The precedence inversion is RFC 6265 5.3-CORRECT and matches the deliberate design at :1899. Base was equally wrong (absint('abc')=0), just reached Expires first - so NOT a regression. Impact is a duration label; :747's value is PHP-generated, so only :1808 is realistically reachable.
**Auto-recommendation (medium):** Mirror set_cookie_is_deletion()'s Max-Age validator in the duration computation so a malformed value defers to Expires.

**Evidence:**
- Verified. :1808 enters the max-age branch on any non-empty value; (int)'abc'=0 fails >0 and the elseif at :1813 never runs. No upstream sanitisation - parse_set_cookie (:1959) stores raw attribute text, and scan_page headers are remote-controlled.
- The precedence inversion is RFC 6265 5.3-CORRECT and matches the deliberate design at :1899. Base was equally wrong (absint('abc')=0), just reached Expires first - so NOT a regression. Impact is a duration label; :747's value is PHP-generated, so only :1808 is realistically reachable.

**Approach:** Mirror set_cookie_is_deletion()'s Max-Age validator in the duration computation so a malformed value defers to Expires.

**Files to modify:**
- `admin/modules/scanner/includes/class-controller.php` — Gate the max-age branch at :1808 (and :747) on the same ^-?[0-9]+$ regex used at :1907, so a non-numeric value falls through to the Expires branch. (why: Two sibling helpers currently disagree about what a valid Max-Age is; the PR added the validator to one of them only.)

**Verification:**
- Feed 'Max-Age=abc; Expires=<two years>' through scan_page() and assert the duration is two years, not 'session'.

**Edge cases to preserve:**
- A valid Max-Age must keep precedence over Expires - that ordering is RFC 6265 5.3-correct and deliberate.

**Latest fix attempt (fixrun_20260820T011724Z3e2e4e):** fixed and verified
- Phase 9 finding: Resolved. Both sites (:749, :1814) now trim and gate Max-Age on the same ^-?[0-9]+$ used by set_cookie_is_deletion(). 9a verified in situ incl. the negative-Max-Age path. No comment calls it a regression.

#### F006 — The client's normalizeUrl() was not updated to match the new PHP normalize_url() rule, so the two now disagree. discover returns '/wp-sitemap.xml' unslashed, the client re-appends the slash, crawls and reports the slashed variant, and the replay queues the slashed variant — the paths this PR stopped rewriting server-side are still rewritten before anything fetches them, and the capture keys diverge between the two halves.

**File:** `admin/assets/js/modules/scan-engine.js:145-152`
**Score:** 58 (moderate)
**Reason:** CONFIRMED: Client normalizeUrl (:145-151) still forces a trailing slash and IS applied to server-supplied URLs via deduplicateUrls(result.urls / result.priority_urls) at :411-414; the slashed form is crawled, pushed to metrics.scannedUrls (:753), sent as scanned_urls (:563) and re-queued by sanitize_scanned_urls (class-controller.php:1170). PHP's new heuristic (:1407-1425) leaves file-like paths unslashed, so the divergence is real and the fix is undone on the browser path. || Scope correction: My /wp-sitemap.xml example was WRONG: discover_urls uses discover_pages_from_db/get_priority_urls (permalinks only), never the sitemap. Real trigger is a file-extension permalink structure such as /%postname%.html.
**Auto-recommendation (high):** Align the client's normalizeUrl() with the PHP rule the PR introduced.

**Evidence:**
- Client normalizeUrl (:145-151) still forces a trailing slash and IS applied to server-supplied URLs via deduplicateUrls(result.urls / result.priority_urls) at :411-414; the slashed form is crawled, pushed to metrics.scannedUrls (:753), sent as scanned_urls (:563) and re-queued by sanitize_scanned_urls (class-controller.php:1170). PHP's new heuristic (:1407-1425) leaves file-like paths unslashed, so the divergence is real and the fix is undone on the browser path.
- My /wp-sitemap.xml example was WRONG: discover_urls uses discover_pages_from_db/get_priority_urls (permalinks only), never the sitemap. Real trigger is a file-extension permalink structure such as /%postname%.html.

**Approach:** Align the client's normalizeUrl() with the PHP rule the PR introduced.

**Files to modify:**
- `admin/assets/js/modules/scan-engine.js` — Apply the same file-like-path exemption at :145-152 that class-controller.php:1407-1425 now uses, so the client stops re-appending a trailing slash to paths the server deliberately left bare. (why: deduplicateUrls() applies the client rule to server-supplied URLs at :411-414, so the browser path undoes the server-side fix before anything is fetched.)

**Verification:**
- On a /%postname%.html permalink structure, assert the URL crawled and reported matches the URL discover returned, byte for byte.

**Edge cases to preserve:**
- Directory paths must still be slashed; /.well-known/ must stay a directory.

**Latest fix attempt (fixrun_20260820T011724Z3e2e4e):** fixed and verified
- Phase 9 finding: Resolved. Client normalizeUrl() mirrors the PHP rule incl. the leading-dot exclusion. Equal on all 13 fixture cases in both runners; 9a mutation-tested both.

#### F007 — home_url( $request_path ) double-prefixes the subdirectory on a WordPress installed in one: REQUEST_URI's path already contains '/blog', so '/blog/shop/x' becomes 'https://site/blog/blog/shop/x'. set_cookie_scope() then derives a wrong RFC 6265 default-path, which is persisted and baked into set_cookie_identity(). Both the writer (:305) and the reader (:723) are wrong in the same direction, which is what hides it — but an explicitly-pathed cookie compared against a default-pathed one no longer matches, so an ordered clearing directive can fail to cancel its prior sighting.

**File:** `admin/modules/scanner/includes/class-controller.php:316-320`
**Score:** 60 (moderate)
**Reason:** CONFIRMED: Confirmed. wp-includes/link-template.php:3481-3483 appends the path to get_option('home') VERBATIM, so on a subdirectory install home_url('/blog/shop/x') yields '.../blog/blog/shop/x'. class-controller.php:316-320 and :722-724 feed exactly REQUEST_URI's path, and the doubled URL reaches set_cookie_scope()'s RFC 6265 default-path branch (:1867-1875), persisted at :360-361 and keyed by set_cookie_identity(). Both callsites are NEW in this PR. || Scope correction: Host unaffected; only the default-path branch (no explicit Path) is wrong, and only on subdirectory installs. Read/write use the same wrong URL, so damage is mismatching explicit-Path vs default-Path rows plus the stored path.
**Auto-recommendation (high):** Build the observation URL from the site root plus the request path, without re-prefixing the home path.

**Evidence:**
- Confirmed. wp-includes/link-template.php:3481-3483 appends the path to get_option('home') VERBATIM, so on a subdirectory install home_url('/blog/shop/x') yields '.../blog/blog/shop/x'. class-controller.php:316-320 and :722-724 feed exactly REQUEST_URI's path, and the doubled URL reaches set_cookie_scope()'s RFC 6265 default-path branch (:1867-1875), persisted at :360-361 and keyed by set_cookie_identity(). Both callsites are NEW in this PR.
- Host unaffected; only the default-path branch (no explicit Path) is wrong, and only on subdirectory installs. Read/write use the same wrong URL, so damage is mismatching explicit-Path vs default-Path rows plus the stored path.

**Approach:** Build the observation URL from the site root plus the request path, without re-prefixing the home path.

**Files to modify:**
- `admin/modules/scanner/includes/class-controller.php` — At :316-320 and :722-724 stop passing REQUEST_URI's full path to home_url(); that path already contains the subdirectory, which home_url() prepends again. (why: WP core link-template.php:3481-3483 appends the argument to get_option('home') verbatim, so a subdirectory install gets /blog/blog/... and set_cookie_scope() derives a wrong RFC 6265 default-path that is persisted and keyed.)

**Verification:**
- On a subdirectory install, assert the stored path for a default-path cookie equals the real request directory.

**Edge cases to preserve:**
- Root installs must stay byte-identical.
- The host component is already correct - do not change it.

**Latest fix attempt (fixrun_20260820T011724Z3e2e4e):** fixed and verified
- Phase 9 finding: Resolved. observation_url() builds from the site root. THREE callsites converted (:301, :321, :723) - the third was not in the finding but keys the same persisted rows. Root installs byte-identical, host and port preserved.

</details>

| F014 | 48 | `admin/modules/scanner/includes/class-controller.php:732-745` | collect_browser_scan_session() computes a domain+path-aware identity for deletion handling and then dedupes the surviving set by cookie NAME alone, so two genuinely distinct cookies sharing a name across domains or paths collapse into one row and one of them never reaches the declaration. |  |

## Fix runs

### Run `fixrun_20260820T011724Z3e2e4e` — 2026-08-20T01:40:55Z

- Outcomes: 6 fixed and verified
- Commits: `a825a90`, `d2396cd`

| Finding | Group | Outcome | phase_9_finding |
|---------|-------|---------|-----------------|
| F001 | FG-3 | ✓ fixed and verified | Resolved by explicit user decision to implement the hold rather than correct the commit message. The import-stage catch now holds instead of aborting (discover/iframe still abort). Crucially the hold alone would have LOCKED THE ADMINISTRATOR OUT for the capture TTL - which is exactly what rejectAndAbort's comment says it existed to prevent - so a session is now marked state='held' and start_browser_scan_session() reclaims a held session while still 409-ing a live one. scan-engine.test.mjs check 30, which asserted the abort, was deliberately inverted; check 28 untouched. |
| F003 | FG-1 | ✓ fixed and verified | Resolved. set_cookie_is_deletion() gained an optional $reference_time; persisted-row callers (:303, :725) pass observation_reference_time(), live-header callers (:331, :1740, :1785) keep time(). Phase 9a audited all five callsites. The duration computation at :757 was ALSO moved onto the same reference after 9a flagged it - fixing the deletion check alone would have let the surviving cookie be reported as 'session'. |
| F004 | FG-1 | ✓ fixed and verified | Resolved. Both sites (:749, :1814) now trim and gate Max-Age on the same ^-?[0-9]+$ used by set_cookie_is_deletion(). 9a verified in situ incl. the negative-Max-Age path. No comment calls it a regression. |
| F005 | FG-3 | ✓ fixed and verified | Resolved. Rather than narrowing the retry predicate (the validator's advice: excluding status 0 would forfeit real recovery), the import is now idempotent. The outcome is remembered against the scan id and a resubmit is answered with it, marked duplicate, with no second save and no second background replay. Write order is load-bearing and pinned by a mutation-proved test: remember BEFORE finish, so a request dying between them still answers a resubmit truthfully. |
| F006 | FG-2 | ✓ fixed and verified | Resolved. Client normalizeUrl() mirrors the PHP rule incl. the leading-dot exclusion. Equal on all 13 fixture cases in both runners; 9a mutation-tested both. |
| F007 | FG-1 | ✓ fixed and verified | Resolved. observation_url() builds from the site root. THREE callsites converted (:301, :321, :723) - the third was not in the finding but keys the same persisted rows. Root installs byte-identical, host and port preserved. |

---

fabiodalez-dev:
Superseded by #242, which shipped as v1.27.0. The integration branch carried this work plus the conflict resolutions and the tests that only exist once the three PRs are combined — merging them separately would have lost those.

PR #247, opened 2026-08-23
fix(geo): CCPA banners never rendered for anyone — plus the preset, router and admin-notice gaps behind it

fabiodalez-dev:
Stacked on #244 so the diff here is only the geo/compliance work. Merge #244 first, or retarget this to main once it lands.

All of this came out of investigating 17 long-red E2E specs. They were red on `main` too, so nobody was reading them any more — and four real product defects were hiding behind that noise.

## The one that matters

`Geo_Runtime::model_to_law()` compared the ruleset model for exact equality with `'opt-out'`. **Not one ruleset in the shipped catalogue uses that string.** All 19 US state laws declare `opt-out-with-sensitive-opt-in`, which is the accurate description of CPRA and its siblings — opt out of sale/sharing, opt *in* for sensitive data.

```
  24x  opt-in
  19x  opt-out-with-sensitive-opt-in
   4x  hybrid
```

So the method could never return `ccpa` for any input the plugin actually ships. Every US visitor was classified as opt-in; `load_banner()` then saw the resolved law disagree with a CCPA banner, found no GDPR banner to substitute, and took its fail-closed branch — which suppresses the UI on purpose so an opt-out notice is never shown under an opt-in regime. That branch was working exactly as designed. It was being fed the wrong law.

**A CCPA banner rendered for nobody, in any state, with no error and no log line. This shipped.**

Measured, banner in CCPA mode, browser UA:

```
visitor US/CA, exact match   -> 0 banners   (ruleset ccpa-california, law gdpr)
visitor US/CA, prefix match  -> 1 banner    (ruleset ccpa-california, law ccpa)
visitor IT,    prefix match  -> 0 banners   (correct: opt-in regime)
```

Mutation-proven. `hybrid` deliberately stays on the `gdpr` side — the more protective reading.

## The rest

**Admin country override was ignored for the whole request.** `Geo_Detector::detect()` read its cache before evaluating the override, and the cache is keyed on the IP hash alone, so anything that resolved geo earlier in the request poisoned it. Adding the filter changed nothing until `wp_cache_flush()`.

**Presets declared a dark pattern.** Four of five paired a solid Accept with a transparent Reject, in the notice *and* the preference centre. EDPB 03/2022 and the Garante's June 2021 guidelines both require equal prominence. The plugin already enforces it — `apply_ui_requirements()` copies Accept's styles onto Reject whenever `equal_weight_buttons` is set, and 47 of 48 rulesets set it. So the asymmetric values never reached a visitor: dead data that only misled whoever read the preset. This is also why `pr61-regression:290` failed — it asserted the raw preset value, i.e. the non-compliant styling the runtime overrides.

**`us-router` never existed.** `_index.json` mapped country US to it; it is the only one of 48 references with no file, dating to the original scaffold. Resolution fell through to the catalogue fallback, which happens to be what the index's own comment says it wants. Now pointed at `gdpr-strict` explicitly — same behaviour, now a decision. `US-CA` still resolves through `_us_regions`.

**An opt-out-only install now says so.** When every active banner is CCPA and none is opt-in, European visitors get no banner and nothing explained why. The source comment for that branch says "a correctly-configured install never reaches this branch"; the admin screens never told anyone.

## Test-suite fixes

Three `gdpr.json` defaults had drifted to `false` on the test site with nothing to restore them (`settings`, `preferenceCenter`, `auditTable`). The last one gates *every* accordion listener, so the preference centre rendered accordions that nothing was bound to. The CCPA specs also browsed as default visitors — country alone is not enough, `US` without a state falls back to an opt-in regime.

## Result

17 red specs → 2, both expected: `flyingpress-cache-purge` needs FlyingPress actually caching (it is active here but emits no cache header), and `release-verify-guard-standdown` is flaky on the base commit too.

i18n is resynced (1550 → 1553 msgids) and it_IT is fully translated for the first time in a while — 10 untranslated → 0.


fabiodalez-dev:
Superseded — this work is on `main` and shipped in 1.28.0.

Verified: the `model_to_law()` prefix match is in place (`0 === strpos( $model, 'opt-out' )`), and the Croatian catalogue is renamed to `hr`. On this branch's own key files `main` is now *ahead*, not behind — comparing `class-i18n-helpers.php`, `class-geo-runtime.php` and `class-geo-detector.php` shows `main` holding content the branch lacks.

It was also stacked on #244 rather than on `main`, so it could never have merged on its own. Closing; the branch stays.

PR #248, opened 2026-08-24
fix(admin): stop the banner page offering switches the runtime overrides

fabiodalez-dev:
## Why

Three support reports in a week turned out to be one defect wearing different names: **a control that can be moved, saved, and have no effect, with nothing on the page saying so.** Cache Compatibility Mode, Show Reject Button and Show revisit widget were each fixed as they arrived. Rather than wait for the fourth report, I went looking for the rest.

The search was mechanical, not a guess: every config path that `Geo_Runtime::apply_ui_requirements()` rewrites, checked against every admin control bound to that path.

| The runtime forces | Rule sets | Admin control | Before |
|---|---|---|---|
| `notice…reject.status` | **47/47** | Show Reject Button | fixed previously |
| `revisitConsent.status` | **47/47** | Show revisit widget | fixed previously |
| `notice…accept.status` | **47/47** | Show Accept Button | **lied** |
| Reject colours ← copied from Accept | **47/47** | 3 colour fields | **lied silently** |
| `notice…settings.status`, `preferenceCenter.*` | 26/47 | Show Settings Button | unexplained |
| `behaviours.respectGPC.status` | 24/47 | Respect GPC | unexplained |

## The design decision

**The treatment is deliberately not uniform.** Disabling a control asserts *"this never does anything"* — true only when every rule set carries the requirement.

- **All 47 → disable and explain.** Accept and Reject. The switch cannot take effect for any visitor; leaving it clickable is a plain untruth, and that is what reached the forum.
- **Some → a sentence, and stay usable.** GPC (24/47) and the sensitive-data opt-in (26/47) still govern every visitor *outside* those jurisdictions. Locking them would be an untruth in the other direction: removing a working setting to make a partial case look tidy.
- **Reject colours → a sentence, no lock.** Not a switch at all. Equal prominence paints Reject with the Accept button's styles, so those fields save correctly and render as something else. Disabling them would be wrong — the colours are real, they simply come from Accept.

## The number stops being hand-written

Both existing notes carried the count as a typed string. I got that count wrong **twice today, in opposite directions**, because the rulesets directory also holds `_index.json` and `index.php` and every manual tally counted the intruders differently. Shipped 1.27.x says "48" and "47 of the 48"; the true figure is **47**, and both requirements cover **all** of them.

New `Ruleset_Loader::requirement_coverage( $paths )` reads it from the catalogue, so the copy cannot drift when a 48th rule set lands. It takes dotted paths and counts a rule set when *any* of them is truthy, which is what an or-requirement like `gpc_honored` / `gpc_required` needs.

## Verification

Probed the **rendered** admin page in both states, not the source:

| | runtime on | runtime off |
|---|---|---|
| toggles disabled | 3 | 0 |
| explanations present | 6 | 0 |
| partial notes | `26 of 47`, `24 of 47` | absent |

The coverage figures were cross-checked against an independent count of the JSON files, which agrees exactly (47/47, 47/47, 26/47, 24/47).

One trap worth recording: my first attempt to switch the runtime off used a mu-plugin filter at priority 99999 and appeared to change nothing. `has_filter()` said the filter was registered, so the obvious reading was that my gate was broken. Dumping `$wp_filter` showed `faz-e2e-audit-lab` pinning `__return_true` at `PHP_INT_MAX` — the fixture outranks any mu-plugin. Had I trusted the identical output, I would have "fixed" working code. The switch is the `faz_e2e_geo_runtime_off` option; the site was restored afterwards and verified clean.

## Notes

- Stacked on #247. Retarget to `main` once that merges.
- **Catalogues resynced on this branch** (`49137b7`, `b69c181`). POT regenerated, all six locales merged, the new strings translated into it_IT only; the other five keep their existing translated counts and take the new entries empty for human translators.
- The resync exposed three pre-existing Italian gaps, all of them mine and older than this branch, so I closed them here. Two strings — the Reject lock note and the Cache Compatibility Mode unavailability note — had **never reached any catalogue**, and shipped in 1.27.x rendering in English everywhere. A third was correctly marked fuzzy by msgmerge because its source had gained a sentence; since `msgfmt` drops fuzzy entries from the `.mo`, leaving it would have rendered that help text in English in the Italian admin, on the very control this work exists to explain. it_IT goes 1536/21/2 → **1539 translated / 20 fuzzy / 0 untranslated**, restoring both the zero and the pre-branch fuzzy count.
- Verified rather than assumed: `msgfmt --check` passes on all six catalogues (the two new notes carry positional `%1$d`/`%2$d`, where a reordering is a runtime warning rather than a typo), each `.mo` recompiles byte-identical, and the eight affected msgids were read back out of the **compiled `.mo` binary** — what WordPress actually loads — not out of the `.po`.


fabiodalez-dev:
Superseded — this work is on `main` and shipped in 1.28.0.

Verified rather than assumed: `Ruleset_Loader::requirement_coverage()` is present, and `admin/views/banner.php` carries all six runtime-override notes (`faz-b-accept-locked`, `faz-b-gpc-note`, `faz-b-settings-note` and the rest). The content reached `main` through the unified PR #250 rather than by merging this branch, so the commits differ but the code does not.

Closing rather than merging: merging now would re-apply changes that are already there, against a base that has moved several releases. The branch stays if anything needs to be recovered from it.

PR #251, opened 2026-08-26
Review-only: the six commits that went straight to main during the 1.28.0 release

fabiodalez-dev:
**Review-only PR. Do not merge — every commit here is already on `main`.**

The 1.28.0 release ended with six commits pushed straight to `main`, so CodeRabbit never saw them: it reviews pull requests. This PR exists purely to put that diff in front of it. `base` is pinned at `d6f4c5d`, the merge commit of #250; `head` is `main`. Close it once reviewed.

Two of the six touch shipped plugin code and are the ones worth the attention. The other four are the release tooling, which failed in four distinct ways during the publication and is now the thing that will run unattended for every future version.

---

### `bbfcaa9` — three defects found minutes before publishing

Found by one question: *does geo-targeting need external lists?*

**The page-cache veto fired where no country could be resolved.** `is_country_dependent_output()` returned true purely because the jurisdiction runtime was on. But the runtime being on does not make output vary by country on an install with no trusted CDN header, no server GeoIP module and no GeoLite2 database — `resolve_for_country()` returns the same `fallback-gdpr-most-protective` ruleset for every visitor, the page is invariant, and the veto cost the publisher their whole page cache for a variation that cannot occur. With enforcement now on by default that reached most shared hosting, so 1.28.0 would have recreated, by its own default, the LiteSpeed and FlyingPress reports it exists to answer.

The new gate keys on a country **source being configured**, not on this visitor having resolved one — with a source present the output is country-dependent even for a visitor it happens to miss. Evaluated per request and never memoised, so installing GeoLite2 or moving behind Cloudflare restores the veto on the next request; anything cached before then carried the most-protective shell, so the stale direction over-blocks. **Please check the predicate for a source I have missed** (`has_country_signal_source()`), and whether the fail-open filter `faz_has_country_signal_source` is the right escape hatch.

**`readme.txt` declared a data flow that does not happen.** The External Services section named `ip-api.com` six times, with a block describing the visitor's IP being sent there for country resolution. There is no such call anywhere in the plugin — two stale docblocks were all that remained. For a privacy plugin, an External Services section that *overstates* what leaves the server is the wrong kind of wrong, and it is the section wp.org reviewers read first.

**The mandatory Playground gate could not test the release.** Its blueprint installs by wp.org **slug**, so run before the SVN commit — as `release.md §5b` required — Playground downloads the *previous* release. Moved to immediately after the commit. Also removed a swallowing `.catch(() => {})` on its admin-menu click: an intercepted click was discarded silently and the run then died 60s later on a heading that was never going to appear, reporting a timeout instead of the interception.

One test went red under the veto change: `FlyingPress remains uncacheable when bootstrap is not explicitly enabled`. It was **under-specified, not wrong** — it never said whether a country source existed, because until now it could not matter. Made explicit, and the new branch covered alongside it. Mutation-proved: removing the gate turns the new case red, source byte-identical after restore.

---

### `d30cd25` — the release write-up hook

Every release now gets an article in the FAZ category on fabiodalez.it. Phase F of `publish-release.sh`, deliberately **after** the SVN commit — an article announcing a release that then failed would have to be retracted. A missing body file does not fail the release; it prints what to do.

Goes over **wp-cli on SSH**, so no REST token: the only credential is the existing SSH key. Idempotent per version — it refuses to post a duplicate, so a retry after a failure is safe. The screenshot script blanks the page *behind* the banner because the test site's front page is a wall of E2E fixture pages.

Worth a look: the `printf '%q'` quoting of the title into the remote `wp post create`, and the `trap` that removes the remote temp dir.

---

### `eaa8cb5`, `6c0c3eb`, `41274a2`, `d83585f` — the release path, which failed four ways

These are the interesting ones for a reviewer who likes self-defeating checks. Publishing 1.28.0 took eight attempts; none of them ever wrote anything partial, and each died further along than the last:

1. **The changelog extractor could not match its own heading.** `awk -v v="## \[${VERSION}\]"` hands awk the *string* `## [1.28.0]`, which awk then compiles as a regex where the brackets are a **character class** — so the pattern meant "`##` followed by one of `1 2 8 0 .`" and matched nothing, on a file whose fifth line is `## [1.28.0] — 2026-08-25`. Replaced with `index($0, hdr) == 1`.
2. **The resume marker blocked the resume.** The script writes `.release-state-X.Y.Z.json` so an interrupted run can continue; the file was untracked, so its own "clean main" preflight rejected the tree on exactly the re-run the marker exists to enable.
3. **`svn ci` ran with `--non-interactive` and no credential.** Worth recording why it was not caught: `svn info` and `svn ls` on this repository succeed **anonymously**, so verifying access with a read proves nothing about write. `SVN_PASSWORD` is now honoured, passed on svn's own stdin via a here-string — off argv, and not consuming the stdin the confirmation gates read from.
4. **`assets/` is checked out at `Depth: empty`.** Versioned, but with none of its children, so `svn up` brought nothing, the screenshot copies landed unversioned, `svn add --force` scheduled them as additions, and the commit died with "already exists / is out of date". I misdiagnosed this twice as leftovers from publishing 1.27.1 the same morning. It was structural and would have repeated every release.

Note for #4: I reached the wrong diagnosis because I checked the directory with `svn status | grep -v '^?'` and read zero lines as clean — while the `?` entries I had filtered out were the whole problem.

---

### Verified before this PR

1076 E2E passed / 0 failed, 133/133 unit suites, Plugin Check 0 errors on the wp.org-shape ZIP, and 1.28.0 is live (wp.org serves it, SVN tag holds 752 files matching the ZIP).

## Summary by CodeRabbit

- **Nuove funzionalità**
  - Aggiunta la pubblicazione automatizzata degli articoli relativi alle release, con supporto per screenshot e modalità di anteprima.
  - Introdotti strumenti per acquisire screenshot del banner e della dashboard.

- **Correzioni**
  - Migliorata la gestione della cache quando non sono disponibili fonti di geolocalizzazione.
  - Rimossa la dipendenza da servizi esterni per la rilevazione del paese: la geolocalizzazione utilizza esclusivamente fonti locali o CDN configurate.
  - Corretto il rilevamento delle sezioni nel changelog durante la pubblicazione delle release.

- **Documentazione**
  - Aggiornate le informazioni su geolocalizzazione, privacy e servizi esterni.


fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
Purpose served. Every one of the five findings is fixed and live.

The Major — `has_country_signal_source()` requiring `CF-IPCountry` on the current request, so a cache warmer's un-vetoed response could be served from cache to a visitor whose header did identify their country — shipped in **1.28.1**, together with the dry-run that wrote to production, the predictable remote temp directory, the silenced `svn up` failure, and the environment-dependent test.

The Major is covered by a case that deliberately sends **no** `CF-IPCountry`, mutation-proved: restoring the header requirement turns it red, and the source is byte-identical after restore.

This PR existed only to put six direct-to-`main` commits in front of a reviewer, which worked — five real findings, one of them user-facing, from code that had reached production without any review. Worth repeating the next time a release needs a hotfix straight to `main`.

PR #257, opened 2026-09-01
Advanced Consent Mode: the analytics category decides its own signal

fabiodalez-dev:
The last finding left open by the compliance audit, and the one I got wrong the
first time.

In Advanced Consent Mode, Consent-Mode-aware Google tags (gtag.js / GA4 / Ads)
are deliberately NOT blocked — class-frontend.php exempts them via
is_gcm_managed_script() so they can send cookieless modeled pings and upgrade on
consent. That exemption removes the script gate, which means the consent SIGNAL
becomes the only control over what GA4 does.

And the signal was pointing the wrong way. analytics_storage is granted when
`analytics` OR `performance` is granted, and both ship in the default category
set. So a visitor who DENIED Analytics and granted Performance had GA4 loaded
(exempt) and told analytics_storage was granted — and it wrote its cookies
against an explicit denial. Nothing else was left to stop it.

## Why this is scoped to advanced mode instead of fixing the merge

I tried fixing the merge itself during the audit, twice, and reverted both:
most-restrictive first, then primary-then-alias. Both broke
gcm-tcf.spec.ts's `custom-performance-only` scenario, and that test is right.

In BASIC mode granted-wins is correct. `performance` is an analytics-class
category, so granting it IS a consent to analytics-class storage; and the tags
under `analytics` are hard-blocked by the script gate regardless of the signal,
so a granted signal cannot let anything through. The two modes genuinely need
different answers from the same merge, which is why no single rule worked.

Keyed on the `analytics` category rather than made most-restrictive, because the
exempted tags are Google's own and `analytics` is the category they belong to.
An install that offers only `performance` is unaffected — with no `analytics`
value in the cookie the existing merge still decides, so nothing is silently
denied.

## Tests

gcm-advanced-mode-analytics.test.mjs drives the real gcm.js through real consent
cookies in BOTH modes: basic keeps granting on performance alone, advanced
honours the analytics decision in either direction, and a performance-only
install is not denied in either. Asserting both is the point — a change that
trades one mode for the other passes any suite that only checks one, which is
exactly how the first two attempts got as far as they did.

Mutation-proved: disabling the advanced-mode branch turns exactly the
denied-analytics assertion red. gcm-tcf.spec.ts still passes (3 passed, 1
skipped), confirming the basic-mode contract is untouched. 139/139 unit suites.

## Summary by CodeRabbit

* **Correzioni**
  * Migliorata la gestione del consenso analytics in modalità Avanzata: il valore esplicito della categoria analytics prevale quando disponibile.
  * Mantenuto il comportamento precedente in modalità Base, incluso il fallback tramite la categoria performance.
  * Aggiunta la gestione del fallback quando la categoria analytics non è offerta.
  * Le modifiche successive al consenso, incluse concessione e revoca dell’accesso analytics, vengono ora rilevate correttamente.


fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
@coderabbitai review

fabiodalez-dev:
On the `playwright.config.ts` timeout finding: confirmed, and it is broader than this PR. Measured it — `gotoResilient()` alone can spend 360 s (3 attempts × 60 s goto + 60 s load) and `completeAdminLogin()` doubles that, against a 90 s budget here and **45 s on `main`**. So the retry loops are unreachable code: the outer budget kills the test before a second attempt can start, and a slow run reports a generic `Test timeout exceeded` instead of the fixture's specific login error.

Not fixing it inside this PR on purpose. Both this branch and #257 changed the same config line (45 → 90), so a `main`-based fix conflicts with both, and neither PR is about the harness. Tracked with the full arithmetic and the shared-deadline fix in #262, to land on its own branch once the queue merges.

Issue #245, opened 2026-08-22
A late scan request can resurrect a torn-down capture session


fabiodalez-dev:
## What happens

A scan-tagged page request that arrives after its crawl has ended can bring a torn-down capture session back to life.

The touch/capture path reads the session transient, decides, then writes — and nothing holds the two halves together. A straggler request that read the transient *before* teardown writes it back *after*, so the session reappears with no crawl behind it. The next scan then meets `faz_browser_scan_in_progress` for a session that no tab is driving, and the administrator waits out the idle TTL for a crawl that finished minutes ago.

## How it showed up

Not from a user report — from a test. During work on #244 the new session spec went red once, immediately after a mutation-restore redeploy: a fresh discover 409'd although teardown had verifiably emptied the transients. Four subsequent runs in isolation were green, as were two full no-retry runs.

That intermittency is the signature: it needs a request already in flight at the moment of teardown, which a fast local crawl produces only occasionally and a slow remote one produces more often.

## Why it was not fixed in #244

The obvious repair — make the read-decide-write atomic, or refuse to write a session the teardown has already closed — sits directly on the heartbeat path. On page-cached sites the heartbeat is the *only* request that reaches PHP during a crawl, so it is the single thing keeping a legitimate long scan alive. Getting the guard slightly wrong there kills real scans on exactly the sites that can least afford it, and trades an occasional stale lock for silently truncated crawls.

That is a worse failure, so the race was recorded rather than patched under time pressure.

## It predates the release

The read-then-write is in the touch/capture path as it stood before 1.27.0. The session panel added in #244 makes the consequence *visible* — the administrator now sees the stale session and can end it with one click instead of needing SSH — but neither causes nor worsens it.

## What a fix has to satisfy

- A teardown must be final: a request that was mid-flight when it ran must not resurrect the session.
- A heartbeat for a session that is genuinely alive must still slide the window, including when it is the only request reaching PHP for minutes at a time.
- No new storage. The session already lives in two transients plus a marker cookie; a third source of truth would be its own bug surface.

A generation counter or a teardown tombstone with a short lifetime both look plausible. Neither has been tested.

## Impact today

Bounded and no longer opaque. Worst case is one stale lock, self-clearing within the idle TTL, and #244's panel lets the administrator end it immediately with a button that says what it discards.

fabiodalez-dev:
Fixed on main in 59572daf, shipping in 1.30.0.

[...]

Issue #243, opened 2026-08-22
Browser scans only ever see the site as an administrator


fabiodalez-dev:
## The problem

A browser scan always runs from a logged-in administrator's session. That is structural — the scanner is an admin screen driving an iframe — but it means the crawl sees the site through one very unrepresentative pair of eyes, and the declaration is built from what those eyes saw.

Two failure directions follow, and they point opposite ways.

**Cookies a visitor gets but the administrator does not.** A logged-in admin often bypasses the paths that set them: page caches serve a different variant, carts and sessions behave differently, and pre-consent third parties may be blocked while the scan runs with blocking disabled. Anything only an anonymous visitor receives can be absent from a scan and therefore absent from the public declaration — the direction that matters for compliance.

**Cookies the administrator gets but a visitor does not.** These are correctly excluded today: `Controller::request_is_admin_context()` classifies observations, and admin-context ones are reported for review rather than declared. `wp-settings-*`, `wordpress_logged_in_*` and `wordpress_test_cookie` are exactly this, and excluding them is right.

The difficulty is that the same mechanism decides both, and it decides from the *request*, not from who would receive the cookie in real use.

## Concrete case from today

On fabiodalez.it the catalogue held `_lscache_vary` (LiteSpeed's cache-vary cookie). Classified as admin-context, so not declared.

That is correct **on that site** — verified: an anonymous visitor receives no `Set-Cookie` at all on the home page, contact page, search results or feed, and only `wordpress_test_cookie` on `wp-login.php`. LiteSpeed varies the cache on the logged-in session, so `_lscache_vary` genuinely never reaches an anonymous visitor there.

It would not be correct on a WooCommerce site, where the same cookie appears for anonymous visitors as soon as a cart exists. Same cookie, same plugin, same classification, opposite correctness — decided by site configuration the scanner cannot see.

## Why this is not a bug report

Nothing here is misbehaving against its specification. `request_is_admin_context()` does what it documents, and the reported-not-declared bucket exists precisely so an administrator can review what was set aside. This is a limit of a design that has one vantage point where it needs two.

## Directions worth weighing

- **Scan a second time without the session.** A logged-out pass over the same URL set would show what an anonymous visitor actually receives, and the difference between the two runs is the interesting artefact. Costs a second crawl and needs care not to double-count.
[...]

PR #265, opened 2026-09-05
fix(scanner): a finished scan can no longer be resurrected (#245), and an accidental cookie classification (#243)

fabiodalez-dev:
Two scanner defects, one of them the session race reported in #245 and still reproducible yesterday.

## #245 — a finished scan could come back to life

`touch_browser_scan_session()` reads the session transient, validates it, and only then writes. A capture request already in flight when the crawl ended wrote its pre-teardown array back afterwards. The session returned with no tab driving it, and the next scan met `faz_browser_scan_in_progress` until the idle TTL expired — an administrator waiting out a crawl that had already finished.

Deleting the transient cannot express *"this was closed"*: absence and "never existed" are the same observation, so a late writer has nothing to detect. The teardown now leaves a 60-second tombstone instead, and the touch path re-reads immediately before writing and abandons the write when it finds one.

The tombstone carries only `faz_closed_at`. Every other reader of these two transients already requires a specific field (`user_id`, `token`, `scan_id`), so a tombstone is rejected naturally and cannot itself block the next scan — which would have reproduced the very symptom being fixed. Checked at all five read sites.

## #243 — a classification that was right by accident

Two tables described `wordpress_logged_in_*` and they disagreed. `Cookie_Database` had it in the "frontend-visible (necessary, shown in banner)" group, i.e. fit for the public declaration; `Frontend::is_wp_internal_cookie()` listed the same prefix as internal and suppressed it. The strict answer won only because all seven visitor-facing surfaces consult the name guard as well.

It is an authentication cookie — a logged-out visitor never receives one — and its sibling `wordpress_sec_*` was already internal. The pair is reunited. No visible change, and the test pins that as an invariant rather than leaving it as a claim.

`wordpress_test_cookie` deliberately stays `necessary`: anonymous visitors really do receive it on `wp-login.php`.

This is one slice of #243, not its closure — see the issue for what the review of the other two directions found.

## Testing

Both fixes are mutation-proven, and both tests needed the proof: the #245 test passed against deliberately broken code **twice** before it was measuring anything.

- It first invoked the tombstone helper directly instead of the real teardown, so it verified that `touch()` reacts to a tombstone — true, and no evidence that the teardown produces one.
- It then fired the teardown on the *first* read, where `touch()` sees the closed session in its opening validation and bails long before reaching the code under test.

Only with the teardown injected between the initial read and the re-read do the assertions discriminate. Reverting either half of the fix now turns them red.

- `tests/unit/test-scan-session-teardown-race-php.php` — 6 checks
- `tests/unit/test-scanner-wp-auth-classification-php.php` — 5 checks
- Neighbouring suites unchanged: visitor-check 59, server-cookie-guard 68, cookie-content-i18n 127, scanner-coverage 56

No frontend JS touched, so no `build:min` needed.

## Summary by CodeRabbit

- **Correzioni**
  - Migliorata la gestione delle sessioni di scansione del browser durante chiusure, rinnovi e riprese simultanei, evitando il ripristino involontario di sessioni terminate.
  - Corretta la classificazione del cookie di autenticazione WordPress `wordpress_logged_in_` come cookie interno di WordPress.

- **Test**
  - Ampliati i controlli automatici su sessioni concorrenti, scadenze, accessi amministrativi, versioni di rilascio e classificazione dei cookie.

### DIFF
--- a/admin/modules/scanner/includes/class-controller.php
+++ b/admin/modules/scanner/includes/class-controller.php
@@ -547,6 +547,40 @@ class Controller {
 	 * @param string $scan_id Client-generated identifier used to isolate retries/tabs.
 	 * @return string|\WP_Error Opaque session token, or a conflict response.
 	 */
+
+	/**
+	 * Seconds a teardown tombstone survives.
+	 *
+	 * Only has to outlive requests that were already in flight when the crawl
+	 * ended. A minute is generous for that and short enough that a stale marker
+	 * cannot obstruct a genuinely new scan for long — and the new-scan path
+	 * clears it explicitly anyway.
+	 */
+	const BROWSER_SCAN_TOMBSTONE_TTL = 60;
+
+	/**
+	 * Close a session so late writers can tell it was closed.
+	 *
+	 * @param string $token   Session token.
+	 * @param int    $user_id Owning user.
+	 * @return void
+	 */
+	private static function tombstone_browser_scan_session( $token, $user_id ) {
+		$marker = array( 'faz_closed_at' => time() );
+		set_transient( self::browser_scan_transient_key( $token ), $marker, self::BROWSER_SCAN_TOMBSTONE_TTL );
+		set_transient( self::browser_scan_active_transient_key( (int) $user_id ), $marker, self::BROWSER_SCAN_TOMBSTONE_TTL );
+	}
+
+	/**
+	 * Whether a transient payload is a teardown tombstone rather than a session.
+	 *
+	 * @param mixed $value Transient payload.
+	 * @return bool
+	 */
+	private static function is_browser_scan_tombstone( $value ) {
+		return is_array( $value ) && ! empty( $value['faz_closed_at'] );
+	}
+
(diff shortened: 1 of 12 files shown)


fabiodalez-dev:
Superseded by #267, which is the union of this PR and its sibling.

The two overlapped: both rewrote `wp-fixture.ts`, `global-setup.ts` and `seed-defaults.ts`, and `rc-1290-acceptance.spec.ts` was byte-identical between them. Merging them in sequence would have reverted the per-attempt login deadline, so keeping both open is a regression waiting to happen rather than redundant work.

#267 keeps the better half of each conflict — the code from the scanner branch (per-attempt deadline, `remaining()` throwing instead of returning a zero Playwright reads as *no timeout*, a previously unbounded call now bounded) and the rationale comments from the login branch, which record the measured cost and the rejected first attempt. The code there is byte-identical to what was validated here; the only delta is documentation, and that is verified rather than asserted.

Closing in favour of it.

Issue #262, opened 2026-09-02
E2E login retries are unreachable: inner timeouts exceed the whole test budget


fabiodalez-dev:
The E2E login retries cannot run: the inner timeouts are larger than the whole test budget.

Raised by review on both #256 and #257, independently, against `tests/e2e/playwright.config.ts`. It is not specific to either branch — it is pre-existing and `main` is the worst case.

## The arithmetic

| where | value |
|---|---|
| `playwright.config.ts` test budget, `main` | **45 s** |
| same, on #256 and #257 | 90 s |
| `gotoResilient()` — 3 attempts × (`page.goto` 60 s + `waitForLoadState` 60 s) | up to **360 s** |
| `attemptAdminLogin()` — `#user_login` visible 20 s + `waitForNavigation` 60 s + URL assert 20 s | up to 100 s |
| `completeAdminLogin()` — 2 attempts of the above | up to **920 s** |

A single `page.goto` is allowed 60 seconds against a 45-second test. So on `main` the very first navigation can outlive the entire test, and the 3-attempt loop inside `gotoResilient` plus the 2-attempt loop inside `completeAdminLogin` are unreachable code: the outer budget kills the test before the second attempt can ever start.

## Why it matters, given none of this is shipped code

It only bites when the site is already slow — which is exactly when the failure needs to be readable. Instead of the fixture's specific login error ("still on wp-login.php after N attempts", with the reauth URL), the runner reports a generic `Test timeout of 45000ms exceeded`, pointing at whatever line happened to be awaiting. The retry code was written to make flaky logins diagnosable and it produces the opposite.

That is the same defect shape as #259: a diagnostic that loses the answer precisely in the case it exists for.

It also means the suite's red count is not trustworthy under load. A spec that fails this way looks identical to a product failure, and I have burned real time on exactly that confusion.

## Fix

Not a bump. Raising the budget to cover 920 s makes a hung run take a quarter of an hour to report.

A shared deadline is the right shape: `completeAdminLogin` computes one deadline from the test's remaining budget and passes it down, and `gotoResilient` stops retrying — and per-attempt timeouts shrink — as the deadline approaches. Retries then genuinely happen (several fast ones fit in the budget) instead of one slow attempt consuming everything.

Concretely:

- per-attempt navigation timeout in the low tens of seconds, not 60 — a local WordPress admin page that has not answered in 15 s is not going to answer at 60
[...]

PR #266, opened 2026-09-05
fix(e2e): give the login a budget, and actually call the baseline reset (#262)

fabiodalez-dev:
Two harness defects found while working out why the full suite showed 14 reds. Neither was a product regression — all three suspect specs pass in isolation — but both make the suite lie about what it measured.

## The login had no budget

`loginAsAdmin()` retried on a fixed per-attempt timeout with no overall ceiling, so a slow or wedged login could burn minutes before failing, and the failure surfaced as whatever assertion came next rather than as a login problem. It now runs against a shared 60-second deadline: each attempt gets what remains, capped at 12 seconds, and the loop is bounded by the deadline rather than by a retry count (`MAX_LOGIN_ATTEMPTS = 4` stays only as a runaway guard). A failure now says so plainly — `admin login failed after N attempt(s) within 60000ms: …`.

This matters beyond tidiness. A login failure mid-suite looks exactly like a product bug in whichever spec inherited the broken session, and that is how it wasted time: I first blamed language contamination, dismissed it because a language-aware spec passed, and only later found the real first cause. In a cascade only the first error informs; everything after it describes the world after the break.

## The baseline reset existed and nothing called it

`resetBaseline()` was written, exported, and invoked by nobody — so every run started from whatever the previous one left behind. `global-setup.ts` now calls it first.

It is deliberately gentle. It normalises `languages.default` to `en` and ensures `en` is in `selected`; it does **not** prune other locales. A first, more aggressive version that forced `selected: ['en']` took the suite from 12 reds to 16 — a reset that destroys state other specs legitimately rely on is worse than no reset.

## Testing

The full suite has not run to completion against this branch: the machine hit a load average of 395 mid-run and nginx started returning upstream timeouts, which produces reds that say nothing about the code. The three specs these changes were meant to stabilise pass in isolation. I would rather say that than present an interrupted run as a green one.

## Summary by CodeRabbit

* **Test**
  * Migliorata l’affidabilità dei test end-to-end, con gestione più robusta dei tentativi di accesso amministrativo e dei tempi massimi.
  * L’ambiente di test viene ora ripristinato automaticamente a una configurazione di base coerente, inclusa la lingua predefinita inglese.
  * Aggiornati i controlli di accettazione per verificare dinamicamente la versione del plugin e la coerenza del relativo tag di stabilità.

### DIFF
--- a/tests/e2e/fixtures/wp-fixture.ts
+++ b/tests/e2e/fixtures/wp-fixture.ts
@@ -48,20 +48,52 @@ const TECHNICAL_COOKIE_RE = [
 
 const isTechnicalCookie = (name: string): boolean => TECHNICAL_COOKIE_RE.some((re) => re.test(name));
 
-async function gotoResilient(page: Page, url: string): Promise<void> {
+/**
+ * How long the whole login is allowed to take, and the slice each attempt gets.
+ *
+ * These used to be independent of the test budget, and that made the retry loops
+ * decorative: one page.goto was allowed 60s plus a 60s load wait — 120s — inside
+ * a 90s test. The outer budget killed the run before a second attempt could
+ * start, so a slow login never retried; it just died, and reported
+ * "Test timeout exceeded" pointing at whatever line happened to be awaiting.
+ *
+ * Measured cost of that: a release-verification run produced 14 red tests across
+ * six specs that have nothing to do with authentication. The first failure was a
+ * login (`#wpadminbar` never appeared); everything after it was a spec operating
+ * on a page it believed was wp-admin — saves that silently did nothing, reads
+ * that returned pre-existing values. Hours to attribute, and nothing wrong with
+ * the product.
+ *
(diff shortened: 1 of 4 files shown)


fabiodalez-dev:
Superseded by #267, which is the union of this PR and its sibling.

The two overlapped: both rewrote `wp-fixture.ts`, `global-setup.ts` and `seed-defaults.ts`, and `rc-1290-acceptance.spec.ts` was byte-identical between them. Merging them in sequence would have reverted the per-attempt login deadline, so keeping both open is a regression waiting to happen rather than redundant work.

#267 keeps the better half of each conflict — the code from the scanner branch (per-attempt deadline, `remaining()` throwing instead of returning a zero Playwright reads as *no timeout*, a previously unbounded call now bounded) and the rationale comments from the login branch, which record the measured cost and the rejected first attempt. The code there is byte-identical to what was validated here; the only delta is documentation, and that is verified rather than asserted.

Closing in favour of it.

PR #267, opened 2026-09-06
fix(scanner): make a scan teardown final (#245, #243), and stop the E2E harness reporting phantom product bugs (#262)

fabiodalez-dev:
Supersedes #265 and #266, which overlapped: both rewrote the same four E2E files, in different and partly incompatible ways. Merging them in sequence would have reverted the per-attempt login deadline. This branch is the union, with the better half of each conflict kept and the reasoning preserved.

## What is in it

**Scanner lifecycle (#245).** A capture request in flight when the crawl ended used to write its pre-teardown session back, resurrecting a scan no tab was driving; the next scan then met `faz_browser_scan_in_progress` until the idle TTL expired.

Deleting the transient cannot express *"this was closed"* — absence and "never existed" are the same observation, so a late writer has nothing to detect. The close is now an independent `<key>_closed` record no renewal can overwrite. Two details are load-bearing:

- Writing the marker *into* the session key — the first shape this took — leaves it open to being clobbered by the very late write it exists to defeat.
- Reads drop the options-cache entries first. WordPress caches option lookups, including negative ones, for the whole request; a re-read inside one request otherwise returns that request's own earlier snapshot, so the guard never observes the concurrent write. A guard that cannot see the race is decorative.

The close no longer touches the per-user index (it could revoke a successor) and no longer clears the marker cookie (a delayed teardown must not unset a marker a newer discover already installed).

**Cookie classification (#243).** `wordpress_logged_in_*` was classified `necessary`, i.e. declarable, while `Frontend::is_wp_internal_cookie()` listed the same prefix as internal. Two tables, one cookie, opposite answers — the strict one won only because all seven visitor-facing surfaces consult it. Correct by accident of ordering rather than by design. `wordpress_test_cookie` deliberately stays `necessary`: anonymous visitors really do receive it on `wp-login.php`.

**Login budget (#262).** The retry loop was decorative — one attempt could outlast the whole test, so a slow login never retried. Cost: 14 reds across six specs unrelated to authentication, the first being a login and every later one a spec operating on a page it believed was wp-admin. Now a shared 60s deadline with 12s attempts; `remaining()` throws when exhausted rather than returning 0, which Playwright reads as *no timeout*.

**Baseline reset.** `resetBaseline()` existed and was called by nobody, so every run inherited the previous one's state — twelve banner-copy reds came from a site still on an Italian default. Now wired into global-setup, after prerequisite validation. Deliberately gentle: forcing English-only was the first attempt and took the suite from 12 reds to 16.

**Session specs.** They counted transient rows; the new design intentionally leaves inert `_closed` markers behind, so that question now reports leaks that are not leaks. They ask the controller whether a session is *active* instead — the condition the assertions were always about.

## Testing

Every guard is mutation-proven, because on this work green tests repeatedly turned out to be measuring nothing. Removing the `_closed` guard turns **12** assertions red, removing the cache-busting read **30**, removing the post-write re-read **6**, restoring the old classification **2**.

- `scan session lifecycle`: 68 checks
- `scanner WP auth classification`: 7 checks
- JS unit suites: 150/150
- `rc-1290-acceptance` version probe: verified end-to-end against faz-test

The full E2E suite has not been run to completion: the machine was saturated during this work (load average 395, homepage at 47s, nginx upstream timeouts), which produces reds that say nothing about the code. That run belongs before any release, per `release.md`, and is not claimed here.

No frontend JS changed, so no `build:min` is required.

## Summary by CodeRabbit

* **Miglioramenti**
  * Rafforzata la gestione delle sessioni di scansione browser, inclusa la protezione da rinnovi concorrenti e sessioni scadute.
  * Migliorata la classificazione dei cookie di autenticazione WordPress.

* **Test**
  * Ampliata la copertura dei flussi di login, configurazione iniziale, ripresa e rilascio delle sessioni.
  * Aggiunti controlli per condizioni di gara, scadenze e coerenza delle versioni di rilascio.
  * Normalizzate automaticamente le impostazioni linguistiche dell’ambiente di test.

### DIFF
--- a/admin/modules/scanner/includes/class-controller.php
+++ b/admin/modules/scanner/includes/class-controller.php
@@ -309,7 +309,7 @@ class Controller {
 		add_action(
 			'shutdown',
 			static function () use ( $token ) {
-				$session = get_transient( self::browser_scan_transient_key( $token ) );
+				$session = self::browser_scan_session_record( $token );
 				$user_id = get_current_user_id();
 				if ( ! is_array( $session ) || empty( $session['user_id'] ) || $user_id !== absint( $session['user_id'] ) ) {
 					return;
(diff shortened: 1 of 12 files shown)


fabiodalez-dev:

This PR contributed the #245 scan-teardown tombstone, the #243 cookie classification, and the E2E harness fixes for the login budget and the unwired baseline reset.

Reopen if #275 does not land.

PR #268, opened 2026-09-06
fix(categories): serve category names and descriptions in the visitor's language

fabiodalez-dev:
Fixes the wordpress.org report [*Category names / description in preference banner always in EN, no way to translate*](https://wordpress.org/support/topic/category-names-description-in-preference-banner-always-in-en-no-way-to-translate/).

Banner copy translated correctly on the reporter's site while category names did not, which makes it look like a language-resolution bug. It is not: three independent causes each produce that symptom on their own.

## Why it happened

**No catalogue shipped for these languages.** `get_translations()` looked for `contents/categories/ru.json`, found nothing, and fell through to `en.json`. Russian and Ukrainian catalogues are added here, complete across all seven slugs.

**English was already stored in the non-English slots.** The editor allowed edits in the default language only, so saving materialised `Necessary` into `ru` and `uk`. A populated slot is not empty, so no fallback fires — the English text is *stored data*, and shipping a catalogue cannot repair an existing install on its own. This is the part that makes the fix non-trivial: it has to undo what the broken editor wrote.

`translate_stored_value()` detects a stored value equal to the shipped English default — compared with tags stripped, so an editor-flattened description still matches — and substitutes the translation. Wording that differs is administrator-authored and left alone. The one case it cannot serve is an admin who deliberately wants the English word in a non-English slot; stored text carries no record of whether it was chosen or inherited, and repairing the common case is the better trade. That limitation is written into the docblock rather than left for someone to rediscover.

**Resolution ran through a gate.** The lookup consulted a catalogue only for languages in a hard-coded list of 41. `ru` and `uk` are in it, so this was not the cause here — but any locale outside that list could never reach a catalogue at all. Resolution is per field now: English base, bundled catalogue over it, downloaded catalogue last. A partial download no longer hides the bundled translation for every other category, and an English value in a downloaded file cannot overwrite a local one.

## Two adjacent defects found while reading the path

- `prepare_json()` ran `json_decode()` over legacy plain-text names. Anything that was not JSON decoded to `null` and **the name was silently wiped**. Non-JSON text is preserved now.
- The catalogue cache tested `! $contents`, so an empty result was recomputed on every request. It tests `false ===` now, and the cache key is versioned so no pre-upgrade payload survives the update.

## The editor

A language selector above the table; switching captures drafts first, so edits in one language survive a visit to another, and saving merges rather than replaces — a language the admin never opened keeps its translation, including one no longer selected.

The English fallback **in the editing box is deliberately gone**. An untranslated field shows empty, which is honest and safe: a stored empty value falls back to the catalogue at render time. Showing English there is exactly how the reported bug was manufactured.

## Testing

- `tests/unit/test-category-i18n-php.php` — 77 assertions, driven through `Banner_Rest::build_categories_payload`, the surface the report was actually about, across three storage shapes (JSON string, array, object) and three languages. Also covers a deselected language that must not be wiped, a partial download, and legacy plain text.
- `tests/unit/js/category-editor-i18n.test.mjs` — 21 checks against the real `cookies.js` in jsdom: the multi-language draft round trip, a failed save retaining drafts, and an empty translation staying empty.
- Full unit gate: 149/149. PHP lint clean. No i18n resync needed — the one new string ("Language") is already in the POT and translated.

Mutation-proven rather than merely green: neutralising the repair fails on *"ru necessary stored English name repaired"*, and restoring the shared editor array fails on *"an unsaved draft must not appear in the category filter list"*.

One assertion is honestly uncovered: the null guard added to `captureCategoryDrafts()` is defensive against a DOM shape `renderCategoryEditor` never produces, so no test exercises it.

## Follow-up, not in scope

`admin/modules/banners/includes/class-banner.php` keeps the same `is_faz_translated()` gate, and no `ru`/`uk` banner content ships. It did not surface here because the reporter typed their banner copy by hand, but a Russian site still gets English banner defaults.

## Summary by CodeRabbit

- **Nuove funzionalità**
  - Aggiunto un selettore di lingua nell’editor delle categorie dei cookie.
  - Ora è possibile modificare nome e descrizione in ogni lingua abilitata.
  - Le bozze vengono mantenute durante il cambio lingua e gli errori di salvataggio consentono di ritentare senza perdere le modifiche.

- **Correzioni**
  - Aggiunte le traduzioni integrate per russo e ucraino.
  - Migliorata la gestione delle traduzioni incomplete, dei valori personalizzati e dei testi esistenti.
  - Corretta la visualizzazione delle traduzioni nelle schermate e nei dati restituiti dall’API.


fabiodalez-dev:

This PR contributed per-language editing of category names and descriptions, the repair of English materialised into non-English slots, and the catalogue cache invalidation.

Reopen if #275 does not land.

PR #269, opened 2026-09-06
fix(banner): ship Russian and Ukrainian copy, and resolve catalogues field by field

fabiodalez-dev:
The follow-up noted on #268. The banner resolver carried the same three causes as the category one; it never surfaced in the [wordpress.org thread](https://wordpress.org/support/topic/category-names-description-in-preference-banner-always-in-en-no-way-to-translate/) only because the reporter had typed their banner copy by hand.

## The three causes

**No catalogue shipped for these languages.** `get_translations()` read `contents/ru.json`, found nothing, fell through to `en.json`. Both catalogues are added here — same key structure as English (asserted, not eyeballed), all 78 leaves translated, the `fazCountdownTimer` span and the `<b>Global Privacy Control</b>` markup preserved verbatim, and the category name quoted inside the long preference-center paragraph matched to the category catalogue so the two surfaces cannot name the same category differently.

**The bundled catalogue sat behind a gate.** It was only consulted when the language appeared in `is_faz_translated()`, a hard-coded list of 41 codes — a site on any other locale could not reach a catalogue however complete. The gate is gone. A language with no file still resolves to English, which is what the gate was standing in for.

**A catalogue was taken whole or not at all.** A downloaded translation covering half the banner silently dropped the bundled wording for the other half, because `if (!$contents)` was already false by then. Resolution is per leaf now: English base, bundled catalogue over it, downloaded last. For a non-English target a value equal to the English default is skipped — downloaded catalogues carry untranslated strings verbatim, and letting one through would overwrite a bundled translation with the English it fell back from. The skip is deliberately not applied to English itself.

## One resolver instead of two that disagreed

`get_law_notice_descriptions()` assembled its own view of the same files, gated differently: it read the bundled catalogue *outside* the gate while its neighbour read it *inside*, and its comment claimed both used "the same gate". Both call one resolver now, so the untouched-default baseline the editor reports cannot disagree with what the frontend renders. That agreement is asserted across four languages rather than assumed.

Also: the cache tested `! $contents`, recomputing an empty result on every request across four JSON reads. It tests `false ===` now, under a versioned key so no pre-upgrade payload survives the update.

## Testing

18 assertions in `tests/unit/test-banner-contents-i18n-php.php`, and mutation-proven rather than merely green — restoring the all-or-nothing overlay turns **3** red, removing the English skip **1**, giving the baseline reader its own resolution **2**.

Full unit gate 148/148, PHP lint clean, and the new catalogues are confirmed to ship (`.distignore` excludes root-level config JSON only).

## Known gap, deliberately not in this PR

`set_contents()` fills missing fields from `get_translations()` at save time. A site that saved its banner while no `ru` catalogue existed therefore has **English stored in the `ru` slot**, and stored text wins over any catalogue — so this PR helps untouched installs and new saves, not those.

That case is common: changing a colour is enough to trigger the save that materialises it. Repairing it means walking the stored content tree and replacing leaves that equal the English default, which is the same heuristic used for categories in #268 but applied to the product's most heavily customised surface. Worth doing, worth deciding on separately.


fabiodalez-dev:

This PR contributed the Russian and Ukrainian banner catalogues and the shared field-by-field resolver, with its own cache invalidation.

Reopen if #275 does not land.

PR #270, opened 2026-09-06
i18n: translate the banner strings that were rendering in English

fabiodalez-dev:
Found while investigating why the cookie banner "doesn't start" on [designpolitie.nl](https://designpolitie.nl/). The banner starts fine — but the site runs FAZ in Dutch and four banner strings were coming out in English. The site is configured correctly; the catalogue was not.

## Two causes, and the second one hides

**Empty `msgstr`.** Nothing reports these usefully. `msgfmt --statistics` counts them among 353 untranslated entries per locale, almost all admin-only and invisible to a visitor, so the number that matters is buried. Scanning by *source reference* instead — entries whose `#:` points at `frontend/`, the cookie-table shortcode or the policy generator — separates what a visitor reads from the rest. There were 35.

**Fuzzy entries, which are worse**, because the `.po` looks translated. `msgfmt` drops them from the `.mo`, so the string falls back to English while the catalogue appears complete — the count says "translated" and the visitor sees English.

Reading what `msgmerge` had actually guessed shows the flag was right to be there:

| msgid | fuzzy guess |
|---|---|
| `Too many consent updates.` | *Filter by consent status.* |
| `Invalid AMP consent instance.` | *Invalidate all consents* |
| `Allow %s cookies` | *All cookies* |

The last one **had lost its placeholder**. Clearing the fuzzy flags in bulk — the obvious quick fix, and what a `msgattrib --clear-fuzzy` sweep would do — would have shipped a `sprintf()` with no `%s`. Every fuzzy entry here is replaced, not un-flagged.

## Coverage

32 strings for `de_DE`, 35 each for `fr_FR`, `hr` and `nl_NL`, 16 for `cs_CZ` — matching exactly what each catalogue was missing. Register follows each catalogue's own established choice: formal everywhere except `nl_NL`, which addresses the visitor as *je*. `it_IT` was already complete and is untouched.

## Verification

Through **WordPress's own gettext reader**, not a local parser — `load_textdomain()` against each rebuilt `.mo`:

```
nl_NL   Voorkeuren opslaan | Analytics-cookies toestaan | Gegevensoverdracht
de_DE   Einstellungen speichern | Analytics-Cookies zulassen | Datenübermittlung
fr_FR   Enregistrer les préférences | Autoriser les cookies Analytics | Transfert de données
hr      Spremi postavke | Dopusti kolačiće: Analytics | Prijenos podataka
cs_CZ   Uložit předvolby | Povolit cookies: Analytics | Přenos dat
```

Every `.mo` is byte-identical to a fresh `msgfmt` of its `.po`; `--check-format` passes on all six catalogues; unit gate 147/147.

## Deliberately not included

45 entries per locale are complete **Cookie Policy templates in Markdown** — multi-paragraph legal documents. A site owner publishes those without reading them, so sweeping a machine translation through is worse than leaving them in English. They need a decision of their own, not a bulk pass.


fabiodalez-dev:

This PR contributed the 35 visitor-facing strings that were rendering in English, including the fuzzy entries msgmerge had guessed wrong.

Reopen if #275 does not land.

PR #271, opened 2026-09-06
feat(policy): Cookie Policy templates in Dutch and Croatian

fabiodalez-dev:
Follow-up to #270. Dutch and Croatian shipped a translated interface and were selectable as a policy language, but **no Cookie Policy template existed for either, in any jurisdiction**. `resolve_template_path()` falls back `lang → native → en`, so those sites generated an English legal document — and nothing in the product said so.

Eight templates: `nl.md` and `hr.md` for GDPR, CCPA/CPRA, LGPD and POPIA.

## Translated from English, not from the existing localizations

The older localized templates are **deliberately abridged**: the German, French, Italian and Spanish CCPA files run ~200 words against 659 in English, having dropped several sections. Translating the fuller English source gives Dutch and Croatian operators the more complete document rather than replicating the abridgement.

Register follows the document type, not the interface: formal *u* in Dutch, where the banner uses *je*. A cookie policy is a legal instrument and Dutch legal drafting is formal — matching the banner would have been consistency in the wrong direction.

## Nothing was needed on the admin side

Worth stating, since it was the other half of the request. The Cookie Policy screen **already** renders the shipped text as each override box's placeholder, and **already** shows a notice when the language falls back to another template. Both behave correctly the moment a template exists: the placeholders now read Dutch and Croatian, and the notice stops appearing.

## The guard, and why it is shaped this way

The gap was invisible — a selectable language, a full gettext catalogue, and a silent English fallback. The test now asserts:

1. **Every language ships for every jurisdiction.** Otherwise the fallback substitutes English for some jurisdictions and not others, with no signal.
2. **No localized template drops a placeholder *type* its English source has.** Losing the only occurrence of `{{COMPANY_EMAIL}}` or `{{COOKIE_CATEGORIES}}` takes the controller's contact or the cookie inventory out of a legal document.
3. **The two faithful translations match the English section count and placeholder multiset exactly.**

Rule 2 holds across the entire existing corpus — zero violations — while leaving the abridged templates their editorial freedom. That shape was not the first attempt: demanding exact parity from everything failed on **eleven** pre-existing files, which would have been a red build reporting a decision someone had already made on purpose. Six of those drop one `{{COMPANY_EMAIL}}`, but each still carries the address twice (controller block and rights section), so no contact is actually lost — only the standalone Contact section.

## Testing

Mutation-proven: deleting `lgpd-brazil/hr.md` turns the completeness assertion red; removing one placeholder from a Dutch template turns two red.

Verified against the real resolver rather than by reading it — `nl` and `hr` resolve to their own file in all four jurisdictions, an unknown language still falls back (`en`, or `pt-BR` for LGPD, its native language) — and the Dutch GDPR policy renders as HTML with the title *Cookiebeleid* and its placeholders intact.

Unit gate 147/147, policy-generator suite 151/151.

## Summary by CodeRabbit

* **Nuove funzionalità**
  * Aggiunti modelli di Cookie Policy in olandese e croato per GDPR, CCPA/CPRA, LGPD e POPIA.
  * Il generatore produce documenti completi nella lingua selezionata, senza ricorrere automaticamente all’inglese.
  * Le schermate di personalizzazione mostrano i testi localizzati come segnaposto per entrambe le lingue.

* **Miglioramenti**
  * Chiarita la distinzione tra richieste verificabili dei consumatori e richieste di opt-out dalla vendita o condivisione dei dati nelle policy CCPA/CPRA.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to FAZ Cookie Manager are documented in this file.
 
 ## [Unreleased]
 
+### Added
+- Cookie Policy templates in Dutch and Croatian for all four jurisdictions (GDPR, CCPA/CPRA, LGPD, POPIA). Both locales already offered a translated interface and were selectable as a policy language, but no template shipped for them, so the generator fell back to English and produced a legal document in the wrong language without saying so. The Cookie Policy screen already shows the shipped text as the placeholder in each override box, so those boxes now show Dutch and Croatian and the "no template ships for this language" notice no longer appears.
+
 ## [1.29.0] - 2026-09-02
 
 ### Fixed
--- a/admin/modules/cookie-policy-generator/includes/class-generator.php
+++ b/admin/modules/cookie-policy-generator/includes/class-generator.php
@@ -49,7 +49,7 @@ class Generator {
 	 *
 	 * @var string[]
 	 */
-	const LANGUAGES = array( 'en', 'it', 'fr', 'de', 'es', 'pt-BR', 'bg', 'cs' );
+	const LANGUAGES = array( 'en', 'it', 'fr', 'de', 'es', 'pt-BR', 'bg', 'cs', 'nl', 'hr' );
 
 	/**
 	 * Canonicalise and validate a policy language code.
--- a/admin/modules/cookie-policy-generator/includes/cookie-policy-gettext-catalog.php
+++ b/admin/modules/cookie-policy-generator/includes/cookie-policy-gettext-catalog.php
@@ -208,7 +208,9 @@ As a California resident, you have the right to:
 - **Right to Non-Discrimination** (§ 1798.125) — we will not discriminate against you for exercising any of these rights
 - **Right to Data Portability** (§ 1798.130(a)(2)) — receive Personal Information in a readily usable format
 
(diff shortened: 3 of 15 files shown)


fabiodalez-dev:
On the review point about the right to limit sensitive personal information: I checked the suggestion against the code before writing it, and it does not hold as proposed.

The suggestion was to say the limit request is submitted through "Manage Preferences". The CCPA ruleset does declare `"model": "opt-out-with-sensitive-opt-in"`, which reads like evidence that such a control exists — but `Geo_Runtime::model_to_law()` only prefix-matches `opt-out` to choose a banner law, and nothing anywhere reads the `-with-sensitive-opt-in` suffix. A search for a sensitive-data control in the preference centre returns nothing: no toggle, no link, no category. Documenting that route would have put a non-existent feature into a legal document, which is worse than the omission, because an omission is discoverable and a false promise is not.

The underlying point was right, though. The rights list granted the right and the text gave no way to exercise it, and a reader would reasonably assume it fell under the verified email path described just above. So the fix names the channel that does exist — email — and extends the no-verification clause to cover both requests, per CPPA Reg. § 7060(b), which bars requiring identity verification for a request to limit just as § 1798.135 does for an opt-out.

The change landed in #275, which consolidates this PR with the other seven, together with the regenerated gettext catalogue and the two CCPA-derived golden renders. This branch is superseded by that one.

fabiodalez-dev:

This PR contributed the Dutch and Croatian Cookie Policy templates, the template-matrix guard, and the CCPA verification wording (extended in #275 to cover the right to limit).

Reopen if #275 does not land.

PR #272, opened 2026-09-06
i18n: complete all five incomplete catalogues (1596/1596 in every locale)

fabiodalez-dev:
Builds on #270. All six shipped locales now sit at **1596/1596** — no untranslated entries, no fuzzy ones. Czech was at 91%, the other four at 75%.

## Why completeness, not tidiness

translate.wordpress.org holds **1,807 strings contributed by volunteers** (747 cs, 534 nl, 526 hr) and wp.org has generated **zero language packs**, because packs are only built above a completeness threshold and the project sits at 27–39% there. That work reaches nobody today.

Complete catalogues can be imported to clear the threshold — after which every future volunteer contribution ships on its own, instead of waiting for one of us to cut a release.

## Existing translations were never in scope

1,713 entries were filled: **1,458 empty and 255 fuzzy**. The extraction handed the translators only those two kinds; the established translations went along **read-only**, as a terminology and register reference. The applier then refuses outright to write into any entry that already carries a real translation, and counts the refusals.

**Fuzzy entries are the apparent exception and are not one.** `msgfmt` drops them from the `.mo`, so they render in English exactly like empty ones — and they are not translations but `msgmerge` guesses copied from *different* strings:

| msgid | fuzzy value |
|---|---|
| `Too many consent updates.` | *Filter by consent status.* |
| `Allow %s cookies` | *All cookies* |

The second had lost its placeholder. Clearing the flags in bulk — the one-command fix — would have shipped a `sprintf()` with no `%s`.

## Three guards, each added after a real failure

Not designed up front; each one exists because something actually went wrong.

1. **Placeholder parity.** A translation whose `%s` / `{{TOKEN}}` / HTML multiset differs from the source is rejected, never written.
2. **`.po` syntax.** Values arrive **already escaped** — `\n` and `\"` are two characters, not a newline and a quote. They are written verbatim. The first version re-escaped them, turning `\n` into a literal backslash-n; `msgfmt` accepts that without a word, so it would have shipped as visible garbage in 1,700 strings. Caught by counting escapes before and after, not by checking the file compiled.
3. **Newline normalisation.** One translator wrote real newlines where the source carries the escape. The msgid states the expected form, so those 44 Cookie Policy sections are repaired mechanically rather than discarded.

A fourth near-miss is worth recording: the syntax guard's own regex matched *two* backslashes instead of one and silently rejected 4 valid Czech translations. It surfaced only because two independent measurements of the same thing disagreed — the applier said "4 rejected", a separate check said "0".

## Plural forms

The applier skips plural entries by design, so the one plural string was written by hand in all four locales — with the **three** forms Croatian requires against two for the others. Its fuzzy value was unrelated to the msgid and had lost its `%d`.

## Verification

Every claim here was measured, not asserted:

- every `.mo` is byte-identical to a fresh `msgfmt` of its `.po`
- `--check-format` passes on all six catalogues
- no value contains a real newline
- the double-backslash count is unchanged from before this work in every file (1 in cs/it, 3 in the rest — all pre-existing)
- the strings resolve through **WordPress' own gettext reader**, `sprintf` included
- unit gate 147/147

Each translator's self-report was also checked rather than believed, and in three cases out of five it was wrong about its own output — the French one reported newlines preserved while it had written real ones; the Croatian one claimed identical escape counts while having deliberately converted 27 entries to Croatian typographic quotes. The work was sound in every case; the reports were not. The French newlines were a genuine defect, and the guard stopped them, not the report.

## The remaining step is manual

Importing to translate.wordpress.org needs PTE rights on the project, which the plugin author can request per locale; GlotPress then accepts these `.po` files directly through *Import Translations*.

## Known, pre-dating this work

German mixes `GDPR` (49) with `DSGVO` (30); Dutch mixes `GDPR` (53) with `AVG` (26). Both catalogues were already inconsistent. The new strings follow each catalogue's own majority rather than imposing a convention across 1,700 entries — worth settling separately, in one deliberate pass.


fabiodalez-dev:

This PR contributed the completion of all five catalogues to 1596/1596.

Reopen if #275 does not land.

PR #273, opened 2026-09-06
refactor(admin-js): name the i18n helper fazI18n, not __

fabiodalez-dev:
Removes the ceiling found while preparing the translate.wordpress.org import in #272.

## The problem, measured

Each admin JS file declares its own helper:

```js
function __( key, fallback ) {
    var parts = key.split('.');
    var obj = (window.fazConfig && window.fazConfig.i18n) || {};
    …
    return typeof obj === 'string' ? obj : fallback;
}
```

It is **not gettext** — it reads a dotted key out of a map PHP already translated. But because it is named `__`, the extractor behind translate.wordpress.org matches it against WordPress' own `__()` signature, where argument one *is* the translatable text, and harvests the keys as if they were English.

Extracting **without domain filtering**, which is evidently what wp.org does:

| tree | entries | dotted keys |
|---|---|---|
| before (HEAD) | **1882** | 267 |
| after (renamed) | **1597** | **0** |

1882 matches the wp.org project's 1881 plus a header exactly; 1597 matches our own catalogue. The 285 difference is the phantom set.

Those entries cannot legitimately be translated, so every locale was capped at **1596/1881 = 84.8%** however complete our catalogues became, and no language pack could be generated. Removing them lets a full import read 100%.

## A local check that says "no change" is measuring the wrong extractor

`wp i18n make-pot` with default settings shows **no difference at all** before and after this change. It filters by text domain, and our second argument is an English fallback rather than a domain, so it discards these calls entirely. That is why the table above uses `--ignore-domain`. Worth stating, because the obvious verification here returns a confident, useless answer.

## No behaviour changes

Keys, English fallbacks and the lookup are untouched; only the function's name differs. `fazConfig.i18n` is unchanged, so PHP needs no edit.

## Verification

- **The diff is the rename and nothing else** — reversing `fazI18n(` → `__(` and dropping the three inserted comment lines reproduces each of the 11 files byte-for-byte against HEAD.
- No `__(` remains anywhere under `admin/assets/js`, in any spacing; `function __` is gone.
- `node --check` passes on all 11 files.
- **All 11 admin screens load with zero console errors and zero ReferenceErrors** — Dashboard, Banner, Cookies, Consent Logs, GCM, Languages, Settings, Setup, GVL, Import/Export, Cookie Policy.
- That probe is **mutation-proven**: renaming a single definition turns the Cookies screen red with `ReferenceError: fazI18n is not defined`, and restoring it turns it green. A page-load check that has never been seen to fail proves nothing about 289 call sites.
- Unit gate 147/147; admin E2E slice (admin-and-api, admin-rest-routes, release-verify-admin-misc, setup-wizard) 39/39.

## What happens next

The effect appears on wp.org only after the next release, when GlotPress re-syncs from the stable ZIP and the 285 entries disappear. The import files already prepared from #272 stay valid — they fill the 1,596 real strings today and will read 100% once the project shrinks.

## Summary by CodeRabbit

* **Miglioramenti**
  * Rinominato l’helper interno per le traduzioni da `__()` a `fazI18n()`.
  * Mantenuti invariati messaggi visualizzati, fallback e comportamento delle funzioni.
  * Migliorata l’identificazione delle stringhe traducibili senza modificare l’esperienza nell’interfaccia amministrativa.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to FAZ Cookie Manager are documented in this file.
 
 ## [Unreleased]
 
+### Changed
+- The admin JavaScript translation helper is named `fazI18n()` instead of `__()`. It never was gettext — it looks up a dotted key in the PHP-provided `fazConfig.i18n` map — but under the gettext name the string extractor on translate.wordpress.org harvested 285 of those keys as if they were translatable English text, which capped every locale's completion at 84.8% and blocked language-pack generation. No runtime behaviour changes: the keys, the English fallbacks and the lookup are untouched.
+
 ## [1.29.0] - 2026-09-02
 
 ### Fixed
--- a/admin/assets/js/modules/scan-engine.js
+++ b/admin/assets/js/modules/scan-engine.js
@@ -23,7 +23,10 @@
 	// i18n helper — looks up fazConfig.i18n.<key> with dot-notation, falls back
 	// to the provided string. Mirrors the per-page helper so the engine carries
 	// no dependency on whichever page happens to load it.
-	function __(key, fallback) {
+	// Not named `__`: this is a key lookup into the PHP-provided fazConfig.i18n
+	// map, not gettext. Under the gettext name, translate.wordpress.org harvests
+	// the dotted keys below as if they were translatable English text.
+	function fazI18n(key, fallback) {
 		var parts = key.split('.');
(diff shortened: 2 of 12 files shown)


fabiodalez-dev:

This PR contributed the fazI18n rename that stops translate.wordpress.org harvesting 285 dotted keys.

Reopen if #275 does not land.

PR #274, opened 2026-09-07
ci: retain GitHub-hosted checks and cancel superseded PR runs

fabiodalez-dev:
The self-hosted migration has been withdrawn at the owner’s request. Quality, Plugin Check and CodeQL now always use standard GitHub-hosted Ubuntu runners. Removed actor-dependent routing and VM-only preparation. This also removes the public self-hosted runner-selection issue. Retain concurrency, timeouts, explicit permissions and all existing WordPress/Plugin Check pins and quality gates. The personal-server CI controller is stopped and disabled; private GitLab backups remain active. Local actionlint and diff checks passed. Fresh checks run on commit a98af03; inspect their results before merging. No plugin release or version change.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -8,6 +8,10 @@ on:
   schedule:
     - cron: '0 8 * * 1'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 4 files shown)


fabiodalez-dev:

This PR contributed the CI changes retaining the GitHub-hosted checks and cancelling superseded PR runs.

Reopen if #275 does not land.

PR #278, opened 2026-09-10
Let an administrator declare what the scan could only set aside (#243)

### LINKED ISSUE #243: Browser scans only ever see the site as an administrator

fabiodalez-dev:
Closes #243.

## The problem

A browser scan runs from one logged-in administrator session and cannot tell whether a cookie it saw there also reaches visitors. LiteSpeed's `_lscache_vary` is the issue's own example: never received by an anonymous visitor on a plain blog, received by every one of them on a shop with a cart. Same cookie, same plugin, opposite correctness, decided by site configuration no crawl can observe.

Two of the three directions the issue weighed already shipped — the anonymous header replay in `run_httponly_check()` with its three-bucket diff, and the name-based `wordpress-internal` known set. This adds the third, and makes the answer take effect.

## Keep the rows, not just the names

The import already reported set-aside names as `jar_only_cookies`, with copy inviting the administrator to "add it manually with Add Cookie" — that is, to retype a domain and a lifetime the scan had measured and then discarded. Retyping evidence is how a judgement never gets made.

The rows are now persisted (bounded, replaced per import, structural WordPress names dropped before they reach the bucket) and offered on the Cookies page with a one-click Declare that carries the measured attributes.

## Splitting the display guard from the shredder guard

`Frontend::is_wp_internal_cookie()` answered two different questions with one list: "never shred this" and "never show this".

For authentication and administrative state both answers are facts — a logged-out visitor structurally cannot receive `wordpress_logged_in_*`. For everything else on that list they are guesses about the site: `_lscache_vary`, `comment_author_*`, `wp-postpass_`, `wpdiscuz_nonce_` and `wordpress_test_cookie` are all set for people who are not logged in at all.

`is_declaration_suppressed()` is therefore a new, separate predicate consulted by the seven visitor-facing surfaces. `is_always_allowed_cookie_name()` and the catalogue category map still consult the unmodified `is_wp_internal_cookie()`.

Default behaviour is byte-identical: every name hidden before is still hidden until an administrator says otherwise about that specific name, and a structural name can never be overridden — re-checked on read as well as on write.

## The correction in the second commit

The first commit claimed declaring "can never arm a deletion". That was true only on the server. Making a cookie visible in the banner put it into the per-category list the **client-side** shredder walks, so rejecting that category would have deleted `_lscache_vary`, `comment_author_*` or a wpDiscuz nonce — the gooloo.de breakage documented in `is_always_allowed_cookie_name()`, re-entering by another door.

The payload now carries `neverDelete`, computed server-side from the unmodified predicate, in all three producers (inline store, category list, banner REST). `_fazRemoveDeadCookies()` honours it and it outranks granular per-service and per-cookie denials; `_fazCleanupRevokedCookies()` walks the jar independently and collects the same protected names.

Three further gaps closed in the same commit: a declaration was written with `discovered=true` so the missed-scan tally could later offer it as stale; promoting a row a previous scan had inserted now preserves its category, translations and measured attributes; accumulated misses are cleared and the Cookie Policy caches invalidated even when no INSERT happened.

## i18n

The POT was already 23 strings behind `main`'s own code — button radius, footer link, CCPA request guidance, category translation, dashboard tracking notes, host cron docs — so 1.30.0 would have shipped that UI untranslatable in all six locales. This is independent of the change here; I verified it by regenerating from a clean `main` checkout.

Regenerated, merged with `--no-fuzzy-matching`, all 33 new strings translated into it_IT — including the rewritten CCPA rights section, whose English had changed under a translation that could no longer match it and which therefore already renders in English on Italian sites today. Every `.mo` byte-matches a fresh `msgfmt` of its `.po`.

## Tests

- `tests/unit/test-set-aside-declaration-php.php` — self-contained PHP assertions on the tiers, the override, the refusals and the payload flag, the last asserted against the real producers by reflection rather than a hand-authored payload.
- `tests/unit/js/granular-cookie-shredder.test.mjs` — the protection holds under whole-category rejection, granular per-service and per-cookie denial, and revocation cleanup.
- `tests/unit/test-scanner-bulk-write-php.php` — exercises the real `save_cookies()`, which is what lets it see `discovered=true` and a skipped INSERT at all.
- `tests/unit/js/cookies-stale-safety.test.mjs` — a promoted row's stale-delete control is retired on refresh.
- `tests/e2e/specs/set-aside-declaration.spec.ts` — before/after visitor payload around the same page, plus the direct-route refusal.

Every protection was mutation-checked: removing any one turns the matching assertions red, and each restore is byte-identical.

Existing guards were updated deliberately rather than loosened — the jar-provenance shape check gained the new read-only handoff and still reddens on a real merge into the imported set.

## Status

Unit suite 161/161. The full E2E suite is running now against `3cf02e42`; I will report the result here before this is merged.

## Summary by CodeRabbit

- **Nuove funzionalità**
  - Aggiunta nell’area amministrativa una barra per esaminare i cookie rilevati dal browser ma non associati a pagine scansionate.
  - È possibile dichiarare singolarmente questi cookie, mantenendo dominio e durata rilevati.
  - I cookie interni dichiarati possono comparire nell’informativa senza essere rimossi automaticamente; i cookie di autenticazione restano esclusi.

- **Correzioni**
  - L’elenco si aggiorna dopo scansioni o nuove dichiarazioni e rimuove gli indicatori obsoleti.
  - I cookie protetti non vengono eliminati durante le pulizie automatiche.


fabiodalez-dev:
## Full E2E suite: green

Run per-spec (one Playwright process per file, one worker, site reset between each) against `3cf02e42`.

**128/128 batches, 0 red, 1166 passed, 4 skipped, 0 flaky.**

The four skips, one by one — a skip is not a pass:

| Spec | Reason | Acceptable |
|---|---|---|
| `gcm-tcf` | IAB TCF not enabled in plugin settings | yes, feature off |
| `pr-regression` | Koko Analytics plugin not active | yes, third party absent |
| `smash-balloon-inert-rule` | Instagram Feed plugin not active | yes, third party absent |
| `playground-compat` | needs `RUN_PLAYGROUND_TEST=1` | **not before a release** |

The last one is a release gate, not a PR gate. It has to run before 1.30.0 ships; it does not block this branch.

## Added after the run

`1b55ce7e` completes the Czech, German, French, Croatian and Dutch catalogues. `bcadb855` had brought all six locales to zero untranslated on 2026-09-06; the POT resync in `58aaffe1` added 34 strings that `msgmerge --no-fuzzy-matching` correctly left untranslated, so those five regressed from that baseline.

Among the 34 is the rewritten CCPA/CPRA rights section, whose English had changed under a translation that could no longer match it — every locale had been falling back to English for that section since the rewrite, silently, because gettext returns the msgid rather than signalling a miss.

All six catalogues are now 1626 translated, 0 fuzzy, 0 untranslated. Placeholder parity checked against every msgid, `{{COMPANY_EMAIL}}` preserved in all six, and each `.mo` byte-matches a fresh `msgfmt` of its `.po`.

That commit touches only `.po` and `.mo`. `.po` is not loaded at runtime and the test site runs `en_US`, so the blast radius is the two specs that deliberately switch locale — `cookie-policy-gettext` (cs_CZ) and `multilingual-edge` (de_DE). Both re-run against the new catalogues: **14/14 passed**. Unit suite re-run: 161/161.

I did not re-run the other 126 batches for a `.mo`-only change, and I am stating that rather than implying the whole suite ran on the tip.

## Review note

German, French and Dutch I am confident in. Czech and Croatian are where my confidence is lowest, particularly the CCPA rights section — it is legal text with statutory citations. Worth a native check before publication.

